PR #16, opened 2018-07-12
Fix usage example in README

messense:


birkenfeld:
This is inconsistent with the name shown above in `Cargo.toml`; one of the two needs fixing.

sehyoc:
@messense Can you update the Cargo.toml as well?

messense:
Updated.

sehyoc:
Thanks for changes.

PR #26, opened 2023-04-25
Extensions from nginx-retry development

avahahn:
### Proposed changes
- minimal bindings for events (`nginx_event_t` and company) [(see example use here)](https://github.com/nginxinc/nginx-retry/blob/35690eb85e56440427874b604fab6c6800ca4026/src/retry.rs#L141)
- getters for request status and subrequest depth
- cycle count increment
- flags as argument to subrequest
- subrequest can accept a custom data [(see example use here)](https://github.com/nginxinc/nginx-retry/blob/35690eb85e56440427874b604fab6c6800ca4026/src/retry.rs#L178)
- do not automatically nuke the subrequest body. that should be up to the caller.

### Testing
- All changes tested with [nginx-retry](github.com/nginxinc/nginx-retry).
- Other modules have not yet been tested for compatibility with these changes.

### DIFF
--- a/src/http/request.rs
+++ b/src/http/request.rs
@@ -192,8 +192,8 @@ impl Request {
         self.0.headers_out.status_line = status_line;
     }
 
-    pub fn get_status_line(&mut self) -> &NgxStr {
-        unsafe {NgxStr::from_ngx_str(self.0.headers_out.status_line)}
+    pub fn get_status_line(&mut self) -> ngx_str_t {
+        self.0.headers_out.status_line.clone()
     }
 
     pub fn get_status(&self) -> HTTPStatus {
@@ -295,6 +295,7 @@ impl Request {
         uri: &str,
         flags: u32,
         module: &ngx_module_t,
+        data: Option<*mut c_void>,
         post_callback: unsafe extern "C" fn(*mut ngx_http_request_t, *mut c_void, ngx_int_t) -> ngx_int_t,
     ) -> Status {
         let uri_ptr = &mut ngx_str_t::from_str(self.0.pool, uri) as *mut _;
@@ -306,7 +307,12 @@ impl Request {
         let post_subreq = sub_ptr as *const ngx_http_post_subrequest_t as *mut ngx_http_post_subrequest_t;
         unsafe {
             (*post_subreq).handler = Some(post_callback);
-            (*post_subreq).data = self.get_module_ctx_ptr(module); // WARN: safety! ensure that ctx is already set
+            if let Some(datum) = data {
+                (*post_subreq).data = datum;
+            } else {
+                // WARN: safety! ensure that ctx is already set
+                (*post_subreq).data = self.get_module_ctx_ptr(module);
+            }
         }
         // -------------
 
--- a/src/http/request.rs
+++ b/src/http/request.rs
@@ -269,10 +269,15 @@ impl Request {
         Status::NGX_DONE
     }
 
-    /// how many subrequests deep is this request?
+    /// how many subrequests are available to make in this request
     /// will return NGX_HTTP_MAX_SUBREQUESTS for a parent request.
     pub fn subrequests_available(&self) -> u32 {
-        self.0.subrequests()
+        // 1 is subtracted because this function was caught returning 1 extra
+        // The return value should be (50, 0), with the parent request returning
+        // NGX_HTTP_MAX_SUBREQUESTS.
+        // See http://nginx.org/en/docs/dev/development_guide.html#http_subrequests
+        // for more information
+        self.0.subrequests() - 1
     }
 
     /// Send a subrequest
--- a/src/http/request.rs
+++ b/src/http/request.rs
@@ -269,6 +269,7 @@ impl Request {
     pub fn subrequest(
         &self,
         uri: &str,
+        flags: u32,
         module: &ngx_module_t,
         post_callback: unsafe extern "C" fn(*mut ngx_http_request_t, *mut c_void, ngx_int_t) -> ngx_int_t,
     ) -> Status {
@@ -293,7 +294,7 @@ impl Request {
                 std::ptr::null_mut(),
                 &mut psr as *mut _,
                 sub_ptr as *mut _,
-                NGX_HTTP_SUBREQUEST_WAITED as _,
+                flags as _,
             )
         };
 
--- a/src/http/request.rs
+++ b/src/http/request.rs
@@ -187,6 +187,15 @@ impl Request {
         self.0.headers_out.status = status.into();
     }
 
+    /// Set HTTP Status Line of response
+    pub fn set_status_line(&mut self, status_line: ngx_str_t) {
+        self.0.headers_out.status_line = status_line;
+    }
+
+    pub fn get_status_line(&mut self) -> &NgxStr {
+        unsafe {NgxStr::from_ngx_str(self.0.headers_out.status_line)}
+    }
+
     pub fn get_status(&self) -> HTTPStatus {
         HTTPStatus(self.0.headers_out.status)
     }
--- a/src/http/request.rs
+++ b/src/http/request.rs
@@ -270,9 +270,9 @@ impl Request {
     }
 
     /// how many subrequests deep is this request?
-    /// will return 0 for a parent request.
-    pub fn subrequest_depth(&self) -> u32 {
-        NGX_HTTP_MAX_SUBREQUESTS - self.0.subrequests()
+    /// will return NGX_HTTP_MAX_SUBREQUESTS for a parent request.
+    pub fn subrequests_available(&self) -> u32 {
+        self.0.subrequests()
     }
 
     /// Send a subrequest
--- a/src/http/request.rs
+++ b/src/http/request.rs
@@ -313,13 +313,6 @@ impl Request {
             )
         };
 
-        // ngx_http_subrequest() ensures that the pointer is no longer null
-        let sr = unsafe { &mut *psr };
-
-        if sr.request_body.is_null() {
-            return Status::NGX_ERROR;
-        }
-        sr.set_header_only(1 as _);
         Status(r)
     }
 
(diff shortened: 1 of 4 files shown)


src/http/request.rs
```diff
@@ -265,11 +270,24 @@ impl Request {
...
+        // See http://nginx.org/en/docs/dev/development_guide.html#http_subrequests
+        // for more information
+        self.0.subrequests() - 1
+    }
+
     /// Send a subrequest
     pub fn subrequest(
         &self,
         uri: &str,
+        flags: u32,
```
ivanitskiy, line 288: can we use enum instead?

#[derive(Debug)]
pub enum HttpSubrequestFlag {
    InMemory,
    Waited,
    Clone,
    Background,
}


then flags.into() would do the thing you would need to implement impl From<HttpSubrequestFlag> for ngx_uint_t { which will convert enum into nginx uint

ivanitskiy:
```rust
#[derive(Debug)]
pub enum NgxHttpSubrequestFlag {
    InMemory,
    Waited,
    Clone,
    Background,
}
// usize
impl From<NgxHttpSubrequestFlag> for ngx_uint_t {
    fn from(val: NgxHttpSubrequestFlag) -> Self {
        match val{
            NgxHttpSubrequestFlag::InMemory => NGX_HTTP_SUBREQUEST_IN_MEMORY as _,
            NgxHttpSubrequestFlag::Waited => NGX_HTTP_SUBREQUEST_WAITED as _,
            NgxHttpSubrequestFlag::Clone => NGX_HTTP_SUBREQUEST_CLONE as _,
            NgxHttpSubrequestFlag::Background => NGX_HTTP_SUBREQUEST_BACKGROUND as _,
        }
    }
}
```

then
```rust
pub fn subrequest(
        &self,
        uri: &str,
        module: &ngx_module_t,
        flag: NgxHttpSubrequestFlag,
        post_callback: unsafe extern "C" fn(*mut ngx_http_request_t, *mut c_void, ngx_int_t) -> ngx_int_t,
    )
// <    ... >
        let r = unsafe {
            ngx_http_subrequest(
                (self as *const Request as *mut Request).cast(),
                uri_ptr,
                std::ptr::null_mut(),
                &mut psr as *mut _,
                sub_ptr as *mut _,
                flag.into(),
            )
        };
// <    ... >
```

src/http/request.rs
```diff
@@ -182,9 +182,14 @@ impl Request {
         unsafe { NgxStr::from_ngx_str((*self.0.headers_in.user_agent).value) }
     }
 
-    /// Set HTTP status of response.
-    pub fn set_status(&mut self, status: HTTPStatus) {
-        self.0.headers_out.status = status.into();
+    pub fn get_status(&self) -> HTTPStatus {
```
ivanitskiy, line 185: seems  like `set_status`  was removed

src/http/request.rs
```diff
@@ -265,11 +270,24 @@ impl Request {
...
+        // 1 is subtracted because this function was caught returning 1 extra
+        // The return value should be (50, 0), with the parent request returning
+        // NGX_HTTP_MAX_SUBREQUESTS.
+        // See http://nginx.org/en/docs/dev/development_guide.html#http_subrequests
+        // for more information
+        self.0.subrequests() - 1
+    }
+
     /// Send a subrequest
     pub fn subrequest(
```
f5yacobucci, line 285: Question to self: Why not support args? If needed later it's an API change. If needed, will add.

src/http/request.rs
f5yacobucci: We need to be able to support expressions like this for modules: `flags |= NGX_HTTP_SUBREQUEST_IN_MEMORY|NGX_HTTP_SUBREQUEST_WAITED;`

I think `bitflags` crate might be the most efficient and easy to use.

```
bitflags! {
  flags SubRequestFlags: u32 {
    const WAITED = 0x01,
    const InMemory = 0x02,
    const Clone = 0x04,
  }
}

let flag = WAITED | InMemory;
assert!(flag.contains(WAITED));
assert!(flag.contains(InMemory));
assert!(flag & Clone).is_empty());
```

f5yacobucci: Although enum variants could work too:
```
enum SubRequestFlag {
  Waited = 0b0001,
  InMemory = 0b0010,
  Clone = 0b0100,
}
flags = Waited | InMemory
```

f5yacobucci: If variants work how I hope, going to update accordingly.

src/http/request.rs
```diff
@@ -182,9 +182,14 @@ impl Request {
...
     }
 
-    /// Set HTTP status of response.
-    pub fn set_status(&mut self, status: HTTPStatus) {
-        self.0.headers_out.status = status.into();
+    pub fn get_status(&self) -> HTTPStatus {
+        HTTPStatus(self.0.headers_out.status)
+    }
+
+    pub fn increment_cycle_count(&mut self) {
```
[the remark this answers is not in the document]
f5yacobucci: I'm questioning if we need this method. I don't see it referenced in this branch or in nginx-retry (the module that inspired this PR).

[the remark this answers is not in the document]
f5yacobucci: Since set_count borrows mutable, I think we need to pass self down the stack mutable as well.

src/http/request.rs
f5yacobucci: +1

src/http/request.rs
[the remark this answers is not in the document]
ivanitskiy: right!

src/http/request.rs
ivanitskiy: TIL: https://stackoverflow.com/questions/43509560/how-to-implement-bitwise-operations-on-a-bitflags-enum

src/http/request.rs
f5yacobucci: @ivanitskiy I notice that internal_redirect, the wrapper for ngx_http_internal_redirect doesn't support `args` either.

i see modules do use this feature, often pulled from the Request itself, but not every case.

can we create a `internal_redirect_with_args` which `internal_redirect` calls - that way `internal_redirect` signature remains stable and we open up args to module writers?

ivanitskiy: good question and a good catch. 

I wrote rust fn `internal_redirect` with an intention to combine "named_location" and "internal redirect" functionally. the use-case for a module i had was to implement something similar to `http_auth_request_module` of nginx config and allow using named and regular locations:
```
http {
    #...
    server {
    #...
        location /private/ {
            auth_request     /auth;
           # auth_request     @named_loc_instead;
        }

        location = /auth {
            internal;
            #<....>
        }
    }
}
```

in my understanding `args` are URI arguments (the one that go after`?`, and look like `foo=bar`. as I understand those are part of URI as well. so I guess we can extract them from a) `location` b) expose in the fn as Optin<&str>

`pub fn internal_redirect(&self, location: &str, args: Option<&str>) -> Status {`

ivanitskiy: another option is to have two rust functions that are "safe" and wraps a C call:

```rust
pub fn named_location(&self, location: &str) -> Status {
        ngx_http_named_location((self as *const Request as *mut Request).cast(), uri_ptr);
  }      
```
and 
```rust
pub fn _internal_redirect(&self, location: &str, args: ???) -> Status {
                ngx_http_internal_redirect(
                    (self as *const Request as *mut Request).cast(),
                    uri_ptr,
                    ??? // std::ptr::null_mut(),
                )
}
```

f5yacobucci: That was my conclusion too, `args` is the query and fragments. Args can be pulled from the URI, but my assumption for the argument was to provide flexibility and mutatable args - I do see a case or two where they are passed in separate from what's stored on the request.

Adding `args` to internal_redirect is a clean approach going forward, but it will break existing users. Are we ok with that seeing it's early? I don't want to make it a habit but, on the other hand, it's still a young project.

ivanitskiy: This is part of the "stabilization" process being pre-v1 versions. 

I can see both ways as a go or a combination.
I can offer this as it offers all variants and a backward compatible change (and maybe need a bit fancier doc to explain the differences/ not sure about fn names too):

```rust
    /// Perform internal redirect without args or named location
    /// see `ngx_internal_redirect` and `ngx_http_named_location`
    pub fn internal_redirect(&self, location: &str) -> Status {
        if location.starts_with('@') {
            self.ngx_http_named_location(location)
        } else {
            self.ngx_internal_redirect(location, "")
        }
    }

    /// Invoke ngx_internal_redirect to perform an internal redirect to a location
    pub fn ngx_internal_redirect(&self, location: &str, args: &str) -> Status {
        assert!(!location.is_empty(), "uri location is empty");
        let uri_ptr = &mut ngx_str_t::from_str(self.0.pool, location) as *mut _;
        let args_ptr = if args.len() > 0 {
            &mut ngx_str_t::from_str(self.0.pool, location) as *mut _
        } else {
            std::ptr::null_mut()
        };

        unsafe {
            Status(ngx_http_internal_redirect(
                (self as *const Request as *mut Request).cast(),
                uri_ptr,
                std::ptr::null_mut(),
            ))
        }
    }

    /// Invoke ngx_http_named_location
    pub fn ngx_http_named_location(&self, location: &str) -> Status {
        assert!(!location.is_empty(), "uri location is empty");
        assert!(location.starts_with('@'), "named location must start with @");
        let uri_ptr = &mut ngx_str_t::from_str(self.0.pool, location) as *mut _;

        unsafe {
            Status(ngx_http_named_location(
                (self as *const Request as *mut Request).cast(),
                uri_ptr,
            ))
        }
    }
```

ivanitskiy: note to me: maybe there is a better way to obtain a raw pointer to a str. currently it copies data and allocates a new ngx_str by `let uri_ptr = &mut ngx_str_t::from_str(self.0.pool, location) as *mut _`

f5yacobucci: good feedback, i'll make some edits today and get an update out.

PR #28, opened 2023-05-20
cargo: make macOS linker flags workaround apply to Apple silicon

alexzorin:
This workaround is required for both Intel and Apple silicon Macs. 

I do not think there is a way to target both triples without duplication.


ivanitskiy:
LGTM

PR #31, opened 2023-06-16
Add Event timers and supporting utilities (+ lint and fmt)

f5yacobucci:
Proposed changes

- minimal bindings for events (nginx_event_t and company) ([see example use here](https://github.com/nginxinc/nginx-retry/blob/35690eb85e56440427874b604fab6c6800ca4026/src/retry.rs#L141))
- getters for request status and subrequest depth
- cycle count increment
- flags as argument to subrequest
- subrequest can accept a custom data ([see example use here](https://github.com/nginxinc/nginx-retry/blob/35690eb85e56440427874b604fab6c6800ca4026/src/retry.rs#L178))
- do not automatically nuke the subrequest body. that should be up to the caller.

Testing
- Original testing against [nginx-retry](https://github.com/nginxinc/ngx-rust/pull/github.com/nginxinc/nginx-retry)

### DIFF
--- a/nginx-sys/src/lib.rs
+++ b/nginx-sys/src/lib.rs
@@ -70,11 +70,9 @@ pub use bindings::*;
 /// let data: &str = "example"; // The string to convert
 /// let ptr = str_to_uchar(pool, data);
 /// ```
-pub fn str_to_uchar(pool: *mut ngx_pool_t, data: &str) -> *mut u_char {
+pub unsafe fn str_to_uchar(pool: *mut ngx_pool_t, data: &str) -> *mut u_char {
     let ptr: *mut u_char = unsafe { ngx_palloc(pool, data.len() as _) as _ };
-    unsafe {
-        copy_nonoverlapping(data.as_ptr(), ptr, data.len());
-    }
+    copy_nonoverlapping(data.as_ptr(), ptr, data.len());
     ptr
 }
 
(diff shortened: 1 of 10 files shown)


ivanitskiy:
@f5yacobucci  Please remove links to not publicly available repositories.

f5yacobucci:
> @f5yacobucci Please remove links to not publicly available repositories.

Can do, can you point me to an example? I'm not seeing one at the moment. Should just be nginx.org dev guide links, but I'll continue to scrub.

nginx-sys/src/lib.rs
```diff
@@ -70,11 +70,9 @@ pub use bindings::*;
 /// let data: &str = "example"; // The string to convert
 /// let ptr = str_to_uchar(pool, data);
 /// ```
-pub fn str_to_uchar(pool: *mut ngx_pool_t, data: &str) -> *mut u_char {
+pub unsafe fn str_to_uchar(pool: *mut ngx_pool_t, data: &str) -> *mut u_char {
```
ivanitskiy, line 73: why is that? when unsafe is used in the body of the fn, it is `safe` to invoke it, otherwise, the caller must use `unsafe` to `unsafe { str_to_uchar(...) }` . it is definitely might break many existing codes, let's keep it back. 
what's the reason?

Issue #30, opened 2023-06-01
Examples on upstream/loadbalancer modules


rikatz:
Hey folks, thanks for this crate.

I'm looking into testing ngx-rust with ingress-nginx, to replace parts of the code that deals with upstream/balancer (today it uses openresty).

First of all, bare in mind I'm almost null in Rust.

While trying to follow the examples and some module development blogs (in C), I couldn't move forward to implement a LB module for it.

I'm a bit lost on where are the limits of C/bindings and Rust code, and was wondering if it is possible to get an example on a loadbalancer module to check where am I doing the mistakes.

Thank you!

f5yacobucci:
The coverage varies right now with how much of the total NGINX API this project supports. There are some known gaps and some unknown gaps. This is a good suggestion for an example though. I'll take a look.

ivanitskiy:
This is an example of a community C module for LB: https://github.com/gnosek/nginx-upstream-fair, but no recent updates though.

rikatz:
Yeap, I was also reading https://www.evanmiller.org/nginx-modules-guide.html and https://hodo.dev/posts/post-17-nginx-custom-upstream-module/ but not gonna lie I'm having some issues to port the idea to ngx-rust :)

f5yacobucci:
Status: I'm significantly through an example based on this work: https://github.com/gabihodoroaga/nginx-upstream-module.

The algorithm may change some for example purposes, but the general structure will remain for cross-referencing. I'm hoping to complete and test it soon.

f5yacobucci:
I opened a WIP PR early to help with your work @rikatz. Hopefully it offers some help before we finalize the structure or annotate it.

rikatz:
Thanks @f5yacobucci it already helps a lot! I will take a look and play with it during the weekend :)

PR #35, opened 2023-06-26
WIP: Issues 30 loadbalancer

f5yacobucci:
WIP checkpoint for load balancer examples.

#30

### DIFF
--- a/examples/upstream.rs
+++ b/examples/upstream.rs
@@ -1,0 +1,103 @@
+use ngx::{
+    core,
+    ffi::{
+        nginx_version, ngx_command_t, ngx_conf_t, ngx_http_module_t, ngx_http_upstream_init_peer_pt,
+        ngx_http_upstream_init_pt, ngx_int_t, ngx_module_t, ngx_uint_t, NGX_CONF_TAKE1, NGX_HTTP_MODULE,
+        NGX_HTTP_SRV_CONF, NGX_HTTP_UPS_CONF, NGX_RS_HTTP_SRV_CONF_OFFSET, NGX_RS_MODULE_SIGNATURE,
+    },
+    http::{HTTPModule, Merge, MergeConfigError},
+    ngx_modules, ngx_null_command, ngx_string,
+};
+use std::os::raw::{c_char, c_void};
+
+#[derive(Debug, Default)]
+struct SrvConfig {
+    max: u32,
+    original_init_upstream: ngx_http_upstream_init_pt,
+    origin_init_peer: ngx_http_upstream_init_peer_pt,
+}
+
+impl Merge for SrvConfig {
+    fn merge(&mut self, _prev: &ModuleConfig) -> Result<(), MergeConfigError> {
+        Ok(())
+    }
+}
+
+#[no_mangle]
+static ngx_http_upstream_custom_ctx: ngx_http_module_t = ngx_http_module_t {
+    preconfiguration: Some(Module::preconfiguration),
+    postconfiguration: Some(Module::postconfiguration),
+    create_main_conf: Some(Module::create_main_conf),
+    init_main_conf: Some(Module::init_main_conf),
+    create_srv_conf: Some(Module::create_srv_conf),
+    merge_srv_conf: Some(Module::merge_srv_conf),
+    create_loc_conf: Some(Module::create_loc_conf),
+    merge_loc_conf: Some(Module::merge_loc_conf),
+};
+
+#[no_mangle]
+static mut ngx_http_upstream_custom_commands: [ngx_command_t; 2] = [
+    ngx_command_t {
+        name: ngx_string!("custom"),
+        type_: (NGX_HTTP_UPS_CONF | NGX_HTTP_SRV_CONF | NGX_CONF_TAKE1) as ngx_uint_t,
+        set: Some(ngx_http_upstream_commands_set_custom),
+        conf: NGX_RS_HTTP_SRV_CONF_OFFSET,
+        offset: 0,
+        post: std::ptr::null_mut(),
+    },
+    ngx_null_command!(),
+];
+
+ngx_modules!(ngx_http_upstream_custom_module);
+
+#[no_mangle]
+pub static mut ngx_http_upstream_custom_module: ngx_module_t = ngx_module_t {
+    ctx_index: ngx_uint_t::max_value(),
+    index: ngx_uint_t::max_value(),
+    name: std::ptr::null_mut(),
+    spare0: 0,
+    spare1: 0,
+    version: nginx_version as ngx_uint_t,
+    signature: NGX_RS_MODULE_SIGNATURE.as_ptr() as *const c_char,
+
+    ctx: &ngx_http_upstream_custom_ctx as *const _ as *mut _,
+    commands: unsafe { &ngx_http_upstream_custom_commands[0] as *const _ as *mut _ },
+    type_: NGX_HTTP_MODULE as ngx_uint_t,
+
+    init_master: None,
+    init_module: None,
+    init_process: None,
+    init_thread: None,
+    exit_thread: None,
+    exit_process: None,
+    exit_master: None,
+
+    spare_hook0: 0,
+    spare_hook1: 0,
+    spare_hook2: 0,
+    spare_hook3: 0,
+    spare_hook4: 0,
+    spare_hook5: 0,
+    spare_hook6: 0,
+    spare_hook7: 0,
+};
+
+#[no_mangle]
+extern "C" fn ngx_http_upstream_commands_set_custom(
+    cf: *mut ngx_conf_t,
+    _cmd: *mut ngx_command_t,
+    conf: *mut c_void,
+) -> *mut c_char {
+}
+
+struct Module;
+
+impl HTTPModule for Module {
+    type MainConf = ();
+    type SrvConf = ModuleConfig;
+    type LocConf = ();
+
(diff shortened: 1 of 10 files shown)


src/http/request.rs
```diff
@@ -111,6 +111,15 @@ impl Request {
         self.0.connection
     }
 
+    /// Pointer to a [`ngx_http_upstream_t`] upstream server object.
+    ///
+    /// [`ngx_http_upstream_t`]: is best described in
+    /// https://nginx.org/en/docs/dev/development_guide.html#http_request
+    /// https://nginx.org/en/docs/dev/development_guide.html#http_load_balancing
+    pub fn upstream(&self) -> *mut ngx_http_upstream_t {
```
ivanitskiy, line 119: Most of the `Request` functionality is somewhat safe and returns "wrapped" types (see pool, status, NgxStr, etc). let's aim that upstream is following this, otherwise, this is just writing C in Rust.

I can offer rename:
`pub fn upstream_ptr(&self) -> *mut ngx_http_upstream_t {`

examples/upstream.rs
```diff
@@ -0,0 +1,325 @@
...
+        (*hcpd).conf = Some(hccf);
+        (*hcpd).upstream = upstream_ptr;
+        (*hcpd).data = (*upstream_ptr).peer.data;
+        (*hcpd).client_connection = request.connection();
+        (*hcpd).original_get_peer = (*upstream_ptr).peer.get;
+        (*hcpd).original_free_peer = (*upstream_ptr).peer.free;
+
+        (*upstream_ptr).peer.data = hcpd as *mut c_void;
+        (*upstream_ptr).peer.get = Some(ngx_http_upstream_get_custom_peer);
+        (*upstream_ptr).peer.free = Some(ngx_http_upstream_free_custom_peer);
```
ivanitskiy, line 195: NIT: potentially it is a good use case to "wrap upstream" and see what we can do with a new "wrapped" upstream rust type... not insisting on the current PR.

examples/upstream.rs
```diff
@@ -0,0 +1,325 @@
...
+            .offset(ngx_http_upstream_custom_module.ctx_index as isize)
+            as *const SrvConfig;
+
+        // FIXME method or macro?
+        // Casting from Rust types to C function pointers might be better as macros for
+        // original_init_peer, free, and get.
+        //
+        // Casting to ngx_http_request_t might also benefit. Alternatively a trait which config and
+        // upstream data use may work as well.
+        let original_init_peer: unsafe extern "C" fn(
```
f5yacobucci, line 172: Change these transmute casts to Option.unwrap wherever possible.

examples/upstream.rs
```diff
@@ -0,0 +1,325 @@
...
+            *mut ngx_http_request_t,
+            *mut ngx_http_upstream_srv_conf_t,
+        ) -> ngx_int_t = unsafe { mem::transmute((*hccf).original_init_peer) };
```
f5yacobucci, line 175: Option.unwrap

examples/upstream.rs
```diff
@@ -0,0 +1,325 @@
...
+    }
+);
+
+#[no_mangle]
+unsafe extern "C" fn ngx_http_upstream_get_custom_peer(pc: *mut ngx_peer_connection_t, data: *mut c_void) -> ngx_int_t {
+    let hcdp: *mut UpstreamPeerData = unsafe { mem::transmute(data) };
+
+    //FIXME log
+
+    let original_get_peer: unsafe extern "C" fn(*mut ngx_peer_connection_t, *mut c_void) -> ngx_int_t =
```
f5yacobucci, line 207: Option.unwrap

examples/upstream.rs
```diff
@@ -0,0 +1,325 @@
...
+
+#[no_mangle]
+unsafe extern "C" fn ngx_http_upstream_free_custom_peer(
+    pc: *mut ngx_peer_connection_t,
+    data: *mut c_void,
+    state: ngx_uint_t,
+) {
+    let hcdp: *mut UpstreamPeerData = unsafe { mem::transmute(data) };
+
+    let original_free_peer: unsafe extern "C" fn(*mut ngx_peer_connection_t, data: *mut c_void, ngx_uint_t) =
```
f5yacobucci, line 225: Option.unwrap

examples/upstream.rs
```diff
@@ -0,0 +1,325 @@
...
+        .srv_conf
+        .offset(ngx_http_upstream_custom_module.ctx_index as isize) as *mut SrvConfig;
+
+    // FIXME: ngx_conf_init_uint_value macro is unavailable
+    if (*hccf).max == u32::MAX {
+        (*hccf).max = 100;
+    }
+
+    //FIXME make a trait and call this as a method on SrvConfig?
+    let init_upstream_ptr: unsafe extern "C" fn(*mut ngx_conf_t, *mut ngx_http_upstream_srv_conf_t) -> ngx_int_t =
```
f5yacobucci, line 251: Option.unwrap

examples/upstream.rs
```diff
@@ -0,0 +1,325 @@
...
+        unsafe { mem::transmute((*hccf).original_init_upstream) };
+    if init_upstream_ptr(cf, us) != Status::NGX_OK.into() {
+        return isize::from(Status::NGX_ERROR);
+    }
+
+    (*hccf).original_init_peer = (*us).peer.init;
+    //(*us).peer.init = Some(ngx_http_upstream_init_custom_peer);
```
f5yacobucci, line 258: Verify this: either uncomment/fix or delete this line.

examples/upstream.rs
```diff
@@ -0,0 +1,325 @@
...
+        //
+        // Casting to ngx_http_request_t might also benefit. Alternatively a trait which config and
+        // upstream data use may work as well.
+        let original_init_peer: unsafe extern "C" fn(
+            *mut ngx_http_request_t,
+            *mut ngx_http_upstream_srv_conf_t,
+        ) -> ngx_int_t = unsafe { mem::transmute((*hccf).original_init_peer) };
+
+        {
+            let r: *mut ngx_http_request_t = unsafe { mem::transmute(&request) };
```
f5yacobucci, line 178: Question to self: can a Request.From trait satisfy this?

src/http/request.rs
f5yacobucci, line 119: Is it more idiomatic to return Option<*mut T> here?

And allows callers to Option.unwrap() and `match` against Some / None?

f5yacobucci: Returning the pointer is a patterned followed from other methods, meaning, it's consistent with access to other ngx_http_request_t fields. But should Option be used for all methods?

f5yacobucci:
Adding the upstream method should increment the MINOR version.

examples/upstream.rs
f5yacobucci: I think that's preferable, to wrap it. I've alluded a little to that in the FIXME's. I'm still thinking through how that should look - it's a little complicated to me ATM.

There are a significant number of structures and functions that accept an ngx_http_upstream_t as an argument.
- upstream peers
- upstream nexts
- upstream headers
- upstream locals
- upstream servers
- upstream state

f5yacobucci: I was leaning towards From trait, but need to experiment a little.

I'm hoping a wrapped upstream can help with the `get` and `free` peer functions. If that requires another peer wrapper I need to take that into consideration, but I'd like to hide the Rust -> C func-ptr transmute casts if possible, a macro could possibly hide the boiler plate too perhaps. Once again, need to experiment a little more.

src/http/conf.rs
```diff
@@ -31,3 +31,35 @@ pub unsafe fn ngx_http_conf_get_module_loc_conf(
...
+    let cf: *const T = (*us).srv_conf.offset(module.ctx_index as isize) as *const T;
+    Some(cf)
+}
+
+/// # Safety
+///
+/// The caller has provided a value `ngx_http_upstream_srv_conf_t. If the `us` argument is null, a
+/// None Option is returned; however, if the `us` internal fields are invalid or the module index
+/// is out of bounds failures may still occur.
+pub unsafe fn ngx_http_conf_upstream_srv_conf_mutable<T>(
```
ivanitskiy, line 56: based on this: https://rust-lang.github.io/api-guidelines/naming.html then it may be renamed:
`ngx_http_conf_upstream_srv_conf_as_mut`

src/http/conf.rs
```diff
@@ -31,3 +31,35 @@ pub unsafe fn ngx_http_conf_get_module_loc_conf(
     let http_conf_ctx = (*cf).ctx as *mut ngx_http_conf_ctx_t;
     *(*http_conf_ctx).loc_conf.add(module.ctx_index) as *mut ngx_http_core_loc_conf_t
 }
+
+/// # Safety
+///
+/// The caller has provided a value `ngx_http_upstream_srv_conf_t. If the `us` argument is null, a
+/// None Option is returned; however, if the `us` internal fields are invalid or the module index
+/// is out of bounds failures may still occur.
+pub unsafe fn ngx_http_conf_upstream_srv_conf_immutable<T>(
```
ivanitskiy, line 40: seems to me based on this https://doc.rust-lang.org/std/primitive.pointer.html we can name as `ngx_http_conf_upstream_srv_conf_as_ptr` or 

some ref: https://doc.rust-lang.org/src/core/ptr/const_ptr.rs.html#1629

f5yacobucci:
Closing and re-opening with split commits and working version.

PR #36, opened 2023-07-06
feat: (1 of 2) Supporting changes for upstream example module.

### LINKED ISSUE #30: Examples on upstream/loadbalancer modules

f5yacobucci:
- functions to get mut and const module context from the ngx_http_upstream_srv_conf_t
- Request from trait
- Request upstream getter
- Request peer init macro

Previous logging changes are moved to #40 
Fixes #30  (replaces PR #35 )

### Checklist

- [X ] I have written my commit messages in the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format. 
- [X ] I have read the [CONTRIBUTING](/CONTRIBUTING.md) doc
- [X ] I have added tests (when possible) that prove my fix is effective or that my feature works
- [X ] I have checked that all unit tests pass after adding my changes
- [X ] I have updated necessary documentation
- [X ] I have rebased my branch onto master
- [X ] I will ensure my PR is targeting the main branch and pulling from my branch from my own fork

### DIFF
--- a/src/http/request.rs
+++ b/src/http/request.rs
@@ -79,6 +79,17 @@ macro_rules! http_variable_get {
 #[repr(transparent)]
 pub struct Request(ngx_http_request_t);
 
+impl<'a> From<&'a Request> for *const ngx_http_request_t {
+    fn from(request: &'a Request) -> Self {
+        &request.0 as *const _
+    }
+}
+impl<'a> From<&'a mut Request> for *mut ngx_http_request_t {
+    fn from(request: &'a mut Request) -> Self {
+        &request.0 as *const _ as *mut _
+    }
+}
+
 impl Request {
     /// Create a [`Request`] from an [`ngx_http_request_t`].
     ///
@@ -102,6 +113,17 @@ impl Request {
     pub fn pool(&self) -> Pool {
         // SAFETY: This request is allocated from `pool`, thus must be a valid pool.
         unsafe { Pool::from_ngx_pool(self.0.pool) }
+    }
+
+    /// Returns the result as an `Option` if it exists, otherwise `None`.
+    ///
+    /// The option wraps a pointer to a [`ngx_http_upstream_t`] upstream server object.
+    ///
+    /// [`ngx_http_upstream_t`]: is best described in
+    /// https://nginx.org/en/docs/dev/development_guide.html#http_request
+    /// https://nginx.org/en/docs/dev/development_guide.html#http_load_balancing
+    pub fn upstream(&self) -> Option<*mut ngx_http_upstream_t> {
+        Some(self.0.upstream)
     }
 
     /// Pointer to a [`ngx_connection_t`] client connection object.
--- a/src/http/request.rs
+++ b/src/http/request.rs
@@ -117,12 +117,16 @@ impl Request {
 
     /// Returns the result as an `Option` if it exists, otherwise `None`.
     ///
-    /// The option wraps a pointer to a [`ngx_http_upstream_t`] upstream server object.
+    /// The option wraps an ngx_http_upstream_t instance, it will be none when the underlying NGINX request
+    /// does not have a pointer to a [`ngx_http_upstream_t`] upstream structure.
     ///
     /// [`ngx_http_upstream_t`]: is best described in
     /// https://nginx.org/en/docs/dev/development_guide.html#http_request
     /// https://nginx.org/en/docs/dev/development_guide.html#http_load_balancing
     pub fn upstream(&self) -> Option<*mut ngx_http_upstream_t> {
+        if self.0.upstream.is_null() {
+            return None;
+        }
         Some(self.0.upstream)
     }
 
--- a/src/http/upstream.rs
+++ b/src/http/upstream.rs
@@ -2,9 +2,15 @@
 ///
 /// Initializes the upstream 'get', 'free', and 'session' callbacks and gives the module writer an
 /// opportunity to set custom data.
+///
+/// This macro will define the NGINX callback type:
+/// `typedef ngx_int_t (*ngx_http_upstream_init_peer_pt)(ngx_http_request_t *r, ngx_http_upstream_srv_conf_t *us)`,
+/// we keep this macro name in-sync with its underlying NGINX type, this callback is required to
+/// initialize your peer.
+///
 /// Load Balancing: <https://nginx.org/en/docs/dev/development_guide.html#http_load_balancing>
 #[macro_export]
-macro_rules! http_upstream_peer_init {
+macro_rules! http_upstream_init_peer_pt {
     ( $name: ident, $handler: expr ) => {
         #[no_mangle]
         extern "C" fn $name(r: *mut ngx_http_request_t, us: *mut ngx_http_upstream_srv_conf_t) -> ngx_int_t {
--- a/src/http/upstream.rs
+++ b/src/http/upstream.rs
@@ -1,0 +1,15 @@
+/// Define a static upstream peer initializer
+///
+/// Initializes the upstream 'get', 'free', and 'session' callbacks and gives the module writer an
+/// opportunity to set custom data.
+/// Load Balancing: <https://nginx.org/en/docs/dev/development_guide.html#http_load_balancing>
+#[macro_export]
+macro_rules! http_upstream_peer_init {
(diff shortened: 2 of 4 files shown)


src/http/upstream.rs
ivanitskiy, line 1: let's wrap ngx_http_upstream as we did for HTTP request:

```
use crate::core::*;
use crate::ffi::*;


/// Wrapper struct for an `ngx_http_upstream_s` pointer, providing methods for working with HTTP upstream.
#[repr(transparent)]
pub struct Upstream(ngx_http_upstream_t);

impl<'a> From<&'a Upstream> for *const ngx_http_upstream_t {
    fn from(request: &'a Upstream) -> Self {
        &request.0 as *const _
    }
}
impl<'a> From<&'a mut Upstream> for *mut ngx_http_upstream_t {
    fn from(request: &'a mut Upstream) -> Self {
        &request.0 as *const _ as *mut _
    }
}

impl Upstream {

    /// Create a [`Upstream`] from an [`ngx_http_upstream_t`].
    ///
    ///
    /// # Safety
    ///
    /// The caller has provided a valid non-null pointer to a valid `ngx_http_upstream_t`
    /// which shares the same representation as `Upstream`.
    pub unsafe fn from_ngx_http_upstream<'a>(r: *mut ngx_http_upstream_t) -> &'a mut Upstream {
        &mut *r.cast::<Upstream>()
    }

    /// other "safer" methods or methods that do a "complex/reparative things
}
```

src/log.rs
```diff
@@ -27,3 +27,26 @@ macro_rules! ngx_log_debug_http {
...
+/// This macro is useful when other masks are necessary or when the request logger is not
+/// conveniently accessible.
+///
+/// See https://nginx.org/en/docs/dev/development_guide.html#logging for details and available
+/// masks.
+#[macro_export]
+macro_rules! ngx_log_debug_mask {
+    ( $mask:expr, $log:expr, $($arg:tt)* ) => {
+        let log_level = unsafe { (*$log).log_level };
+        if log_level & $mask as usize != 0 {
```
ivanitskiy, line 43: Probably, it is a bit hard to understand masks here.


The debug masks are:
```
NGX_LOG_DEBUG_CORE
NGX_LOG_DEBUG_ALLOC
NGX_LOG_DEBUG_MUTEX
NGX_LOG_DEBUG_EVENT
NGX_LOG_DEBUG_HTTP
NGX_LOG_DEBUG_MAIL
NGX_LOG_DEBUG_STREAM
```
and in the generated bindings are:
```rust
pub const NGX_LOG_DEBUG_CORE: u32 = 16;
pub const NGX_LOG_DEBUG_ALLOC: u32 = 32;
pub const NGX_LOG_DEBUG_MUTEX: u32 = 64;
pub const NGX_LOG_DEBUG_EVENT: u32 = 128;
pub const NGX_LOG_DEBUG_HTTP: u32 = 256;
pub const NGX_LOG_DEBUG_MAIL: u32 = 512;
pub const NGX_LOG_DEBUG_STREAM: u32 = 1024;
````
so with that macro it is really easy to make a mistake when using it say mask == 2.
let's think how can we improve this and make it easier/more idiomatic to use it. 

one option is that the mask is an enum...

Comment (ivanitskiy) [reply src/log.rs:43]: another options is not original. we have extra macro_rules:
```
ngx_log_debug_http
ngx_log_debug_core, 
ngx_log_debug_XYZ
```
and these would use internally ngx_log_debug_mask`.
like `ngx_log_debug_http` uses `ngx_log_debug` already

File: src/http/upstream.rs
Comment (f5yacobucci) [reply src/http/upstream.rs:1]: I considered that, but it got awkward because the module only sets values on an internal structure and still calls the peer function with the C raw pointer.

I think it'll make more sense if we wrap ngx_peer_connection_t along with it. I think I have an idea to clean this up.

File: src/http/upstream.rs [hunk 2]
```diff
@@ -0,0 +1,15 @@
+/// Define a static upstream peer initializer
+///
+/// Initializes the upstream 'get', 'free', and 'session' callbacks and gives the module writer an
+/// opportunity to set custom data.
+/// Load Balancing: <https://nginx.org/en/docs/dev/development_guide.html#http_load_balancing>
+#[macro_export]
+macro_rules! http_upstream_peer_init {
```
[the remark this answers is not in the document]
f5yacobucci: For these functions, the upstream modules all use this format: `ngx_http_upstream_init_[MODULE_NAME]_peer`

I kind of followed the typedef for the signature, but accidentally transposed peer and init. We can name it identical to that typedef (minus the ngx, this repo tends to omit that prefix).

I also placed it in upstream.rs because the typedef is in ngx_http_upstream.h, so it's a clue and alignment to the original.

If we were more faithful to the NGINX name it would be `http_upstream_init_peer_pt`. I think keeping this in upstream is a helpful alignment to NGINX code base.

src/log.rs
f5yacobucci: Enums are pretty powerful in Rust, but the logger is still storing log_level as a bit field. So we might need to implement the BitOps traits for them.

It's verbose, but I think the simplest and most foolproof way is a function for each mask. _http, _event, _core, etc. If you're ok with this second approach, I think it'll most easily understood by users.

f5yacobucci: There is a safer enum based version with macro pattern matching in PR #40.

Spazzy757:
Hi,

Is there any timeline on when the upstream module will be available?

f5yacobucci:
> Hi,
> 
> Is there any timeline on when the upstream module will be available?

Working on it now, traveling next week, but working to close this review and then publish the module next.

src/http/upstream.rs
f5yacobucci: Spoke offline: building an Upstream and PeerConnection object aren't the direction we want to go at the moment. Leaving this as is and designing a new approach.

src/http/upstream.rs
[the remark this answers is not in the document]
f5yacobucci: Renamed to reflect the function type it defines.

f5yacobucci:
Backed out of adding Upstream object for now. Current commits build and pass lint without warnings.

src/http/request.rs
```diff
@@ -104,6 +115,21 @@ impl Request {
...
 
+    /// Returns the result as an `Option` if it exists, otherwise `None`.
+    ///
+    /// The option wraps an ngx_http_upstream_t instance, it will be none when the underlying NGINX request
+    /// does not have a pointer to a [`ngx_http_upstream_t`] upstream structure.
+    ///
+    /// [`ngx_http_upstream_t`]: is best described in
+    /// https://nginx.org/en/docs/dev/development_guide.html#http_request
+    /// https://nginx.org/en/docs/dev/development_guide.html#http_load_balancing
+    pub fn upstream(&self) -> Option<*mut ngx_http_upstream_t> {
```
ivanitskiy, line 126: nit: 
totally optionally we can start using some [name conventions](https://rust-lang.github.io/api-guidelines/naming.html)  (I know there is no such convention as `_raw_mut` on the link, but why not?)

`pub fn upstream_raw_mut(&self) -> Option<*mut ngx_http_upstream_t> {`

similarly to `_mut` naming convention, from the rust recommendation::

```
pub struct S {
    first: First,
    second: Second,
}

impl S {
    // Not get_first.
    pub fn first(&self) -> &First {
        &self.first
    }

    // Not get_first_mut, get_mut_first, or mut_first.
    pub fn first_mut(&mut self) -> &mut First {
        &mut self.first
    }
}
```

f5yacobucci: A lot of Request doesn't follow a consistent pattern - I'll present a diff to discuss.

f5yacobucci: This will come in a following commit. We've decided to consistently name all accessors and functions.

PR #37, opened 2023-07-06
feat: (2 of 2) Upstream module example

### LINKED ISSUE #30: Examples on upstream/loadbalancer modules

f5yacobucci:
Upstream example
    
- derived from github.com/gabihodoroaga/nginx-upstream-module and NGINX ngx_http_upstream_keepalive_module.c
    
Simply replaces peer functions with passthrough versions, i.e., proxies
traffic through this module to originally configured upstream functions.
    
Useful as a guide and example, but does not add load balancing
algorithms.

Fixes #30 (replaces PR #35)

### Checklist

- [X ] I have written my commit messages in the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format. 
- [X ] I have read the [CONTRIBUTING](/CONTRIBUTING.md) doc
- [X ] I have added tests (when possible) that prove my fix is effective or that my feature works
- [X ] I have checked that all unit tests pass after adding my changes
- [X ] I have updated necessary documentation
- [X ] I have rebased my branch onto master
- [X ] I will ensure my PR is targeting the main branch and pulling from my branch from my own fork

### DIFF
--- a/examples/upstream.rs
+++ b/examples/upstream.rs
@@ -1,0 +1,356 @@
+/*
+ * This example is based on:
+ * https://github.com/gabihodoroaga/nginx-upstream-module
+ * as well as the NGINX keepalive module: ngx_http_upstream_keepalive_module.c.
+ *
+ * The NGINX authors are grateful to @gabihodoroaga for their contributions
+ * to the community at large.
+ */
+use ngx::{
+    core::{Pool, Status},
+    ffi::{
+        nginx_version, ngx_atoi, ngx_command_t, ngx_conf_log_error, ngx_conf_t, ngx_connection_t,
+        ngx_event_free_peer_pt, ngx_event_get_peer_pt, ngx_http_module_t, ngx_http_request_t,
+        ngx_http_upstream_init_peer_pt, ngx_http_upstream_init_pt, ngx_http_upstream_init_round_robin,
+        ngx_http_upstream_module, ngx_http_upstream_srv_conf_t, ngx_http_upstream_t, ngx_int_t, ngx_module_t,
+        ngx_peer_connection_t, ngx_str_t, ngx_uint_t, NGX_CONF_NOARGS, NGX_CONF_TAKE1, NGX_CONF_UNSET, NGX_ERROR,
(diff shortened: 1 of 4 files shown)


examples/upstream.rs
```diff
@@ -0,0 +1,356 @@
...
+
+    let maybe_conf: Option<*mut SrvConfig> =
+        ngx_http_conf_upstream_srv_conf_mutable(us, &ngx_http_upstream_custom_module);
+    if maybe_conf.is_none() {
+        ngx_conf_log_error(
+            NGX_LOG_EMERG as usize,
+            cf,
+            0,
+            "CUSTOM UPSTREAM no upstream srv_conf".as_bytes().as_ptr() as *const i8,
+        );
```
ivanitskiy, line 253: nit: a good candidate to be a macro for ngx crate
ngx_conf_log_error!(LEVEL, "regular text");
seems like we need to revisit logging and understand it better from a usage point of view. Maybe for a next time.

examples/upstream.rs
```diff
@@ -0,0 +1,356 @@
...
+        return isize::from(Status::NGX_ERROR);
+    }
+    let hccf = maybe_conf.unwrap();
+    // NOTE: ngx_conf_init_uint_value macro is unavailable
+    if (*hccf).max == u32::MAX {
```
ivanitskiy, line 258: nit: may be do a dereference here and then you don't need to do `(*hccf)`/`(*us)`?

examples/upstream.rs
f5yacobucci: i'll convert this to an issue.

examples/upstream.rs
f5yacobucci: are you suggesting something like:
`let hccf = *(maybe_conf.unwrap());`

or something different that i'm not understanding?

f5yacobucci: i was looking at the Deref trait, but doesn't seem appropriate here unless i'm missing something more.

PR #40, opened 2023-07-19
feat: Add debug log mask support

f5yacobucci:
NGINX supports multiple debug masks to customize logging.

We use Rust's type system to represent these as an enum and ease a developers usage. They can call the ngx_log_debug_mask with any of the supported enum elements.

To aid in keeping in sync with NGINX primitives a unit test will fail when the current FIRST and LAST log masks do not align any longer.

### Checklist

- [ X] I have written my commit messages in the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format. 
- [ X] I have read the [CONTRIBUTING](/CONTRIBUTING.md) doc
- [ X] I have added tests (when possible) that prove my fix is effective or that my feature works
- [ X] I have checked that all unit tests pass after adding my changes
- [ X] I have updated necessary documentation
- [ X] I have rebased my branch onto master
- [ X] I will ensure my PR is targeting the main branch and pulling from my branch from my own fork

### DIFF
--- a/src/log.rs
+++ b/src/log.rs
@@ -1,25 +1,11 @@
 /// Utility function to provide typed checking of the mask's field state.
 #[inline(always)]
-fn check_mask(mask: DebugMask, log_level: usize) -> bool {
+pub fn check_mask(mask: DebugMask, log_level: usize) -> bool {
     let mask_bits: u32 = mask.into();
     if log_level & mask_bits as usize == 0 {
         return false;
     }
     true
-}
-
-/// Internal macro, provided to reduce code duplication.
-///
-/// Expects an ngx_log_t and message format template.
-macro_rules! _ngx_log_debug_internal {
-    ( $log:expr, $($arg:tt)* ) => {
-        let level = $crate::ffi::NGX_LOG_DEBUG as $crate::ffi::ngx_uint_t;
-        let fmt = ::std::ffi::CString::new("%s").unwrap();
-        let c_message = ::std::ffi::CString::new(format!($($arg)*)).unwrap();
-        unsafe {
-            $crate::ffi::ngx_log_error_core(level, $log, 0, fmt.as_ptr(), c_message.as_ptr());
-        }
-    }
 }
 
 /// Write to logger at a specified level.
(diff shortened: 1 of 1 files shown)


src/log.rs
```diff
@@ -27,3 +46,136 @@ macro_rules! ngx_log_debug_http {
...
+            DebugMasks::Http => crate::ffi::NGX_LOG_DEBUG_HTTP,
+            DebugMasks::Mail => crate::ffi::NGX_LOG_DEBUG_MAIL,
+            DebugMasks::Stream => crate::ffi::NGX_LOG_DEBUG_STREAM,
+        }
+    }
+}
+
+/// Log with appropriate debug mask.
+///
+/// When the request logger is available `ngx_log_debug_http` can be used for `NGX_LOG_DEBUG_HTTP` masks.
```
ivanitskiy, line 103: seems to me the copy/paste was left as is... 

```suggestion
/// When the request logger is available `ngx_log_debug_mask` can be used for `NGX_LOG_DEBUG_HTTP` masks.
```

src/log.rs
```diff
@@ -1,3 +1,27 @@
+// Utility function to provide typed checking of the mask's field state.
```
ivanitskiy, line 1: nit: in rust docs-string is `///`, `//` - a comment.  (if you render documentation `check_mask` would not have documentation.

f5yacobucci: check_mask is private. Doctest can't access private methods from what I've read.

src/log.rs
```diff
@@ -1,3 +1,27 @@
...
+// Internal macro, provided to reduce code duplication.
+//
+// Expects an ngx_log_t and message format template.
+macro_rules! _ngx_log_debug_internal {
+    ( $log:expr, $($arg:tt)* ) => {
+        let level = $crate::ffi::NGX_LOG_DEBUG as $crate::ffi::ngx_uint_t;
+        let fmt = ::std::ffi::CString::new("%s").unwrap();
+        let c_message = ::std::ffi::CString::new(format!($($arg)*)).unwrap();
+        unsafe {
+            $crate::ffi::ngx_log_error_core(level, $log, 0, fmt.as_ptr(), c_message.as_ptr());
```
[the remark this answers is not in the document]
f5yacobucci: This is what NGINX does as well.

ngx_log_debug checks for a mask. If set, it calls ngx_log_error_core with NGX_LOG_DEBUG.

This internal macro is "_" prefixed and without a macro_export so can't be used outside of this mod.

ngx_log_debug and ngx_log_debug_mask each use this after they've done their specific checks.

src/log.rs
f5yacobucci: This might be poor wording. This is correctly referring to ngx_log_debug_http. Its intent is to say ngx_log_debug_http should be preferred when you can use a Request.

Basically justifying ngx_log_debug_http AND ngx_log_debug_mask (for when you don't have a specific Request to use).

src/log.rs
f5yacobucci: Will private functions still get documentation rendered?

ivanitskiy: anyways, if your intention is to document then probably is best to use `///` over '//'

src/log.rs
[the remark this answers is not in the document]
ivanitskiy: anyways, if you tested it with nginx and it works as expected, then let's merge.

f5yacobucci:
Believe I have everything addressed. Let me know if you're ok and I'll merge.

ivanitskiy (approved):

PR #43, opened 2023-08-28
WIP: multiple changes 

ivanitskiy:
### Proposed changes
This is a temporary PR to share changes. I intend to break it down into a few separated PRs

### Checklist

- [ ] I have read the [CONTRIBUTING](/CONTRIBUTING.md) doc
- [ ] I have checked that all unit tests pass after adding my changes
- [ ] I have rebased my branch onto master
- [ ] I will ensure my PR is targeting the main branch and pulling from my branch from my own fork

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,13 +77,12 @@ We provide a multistage [Dockerfile](Dockerfile):
     docker buildx build --build-arg NGX_VERSION=1.23.3 -t ngx-rust .
 
     # start NGINX using [curl](examples/curl.conf) module example:
-    docker run --rm -d  -p 8000:8000 ngx-rust nginx -c examples/curl.conf
+    docker run --name curl --rm -d  -p 8000:8000 ngx-rust nginx -c examples/curl.conf
 
-    # test it - you should see 403 Forbidden
+    # Test it - you should see 403 Forbidden
     curl http://127.0.0.1:8000 -v -H "user-agent: curl"
 
-
-    # test it - you should see 404 Not Found
+    # Test it - you should see 404 Not Found
     curl http://127.0.0.1:8000 -v -H "user-agent: foo"
 
 ## Usage
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 
 ## Project status
-This project is still a work in progress and not production ready.
+This project is still a work in progress and not production-ready.
 
 # Description
 
@@ -14,15 +14,21 @@ dynamic modules completely in Rust.
 
 In short, this SDK allows writing NGINX modules using the Rust language.
 
+It contains the following Rust crates:
+ * [nginx-sys](./nginx-sys) - allows to use ngx_* C functions via FFI when implementing modules. The `-sys` is used to follow a [naming convention](https://doc.rust-lang.org/cargo/reference/build-scripts.html#-sys-packages) to link with a C library.
+ * [ngx](./src) - it an opinionated SDK to make it a bit easer to use [nginx-sys](./nginx-sys) crate. Implements `macro_rules`, provides a way to build a dynamic module without any C code (see `ngx_modules!` macro_rule).
+
 ## Build
 
 NGINX modules can be built against a particular version of NGINX. The following environment variables can be used to specify a particular version of NGINX or an NGINX dependency:
 
-* `ZLIB_VERSION` (default 1.2.13) -
+* `ZLIB_VERSION` (default 1.3)
 * `PCRE2_VERSION` (default 10.42)
 * `OPENSSL_VERSION` (default 3.0.7)
 * `NGX_VERSION` (default 1.23.3) - NGINX OSS version
 * `NGX_DEBUG` (default to false)-  if set to true, then will compile NGINX `--with-debug` option
+* `NGX_SRC_DIR` (default not set) - When the value is set, then use this NGINX source code folder to build bindings from
+* `NGX_CONFIGURE_ARGS` (default not set) - When the value is set, then run the NGINX configure script with
 
 For example, this is how you would compile the [examples](examples) using a specific version of NGINX and enabling
 debugging:
@@ -30,7 +36,7 @@ debugging:
 NGX_DEBUG=true NGX_VERSION=1.23.0 cargo build --package=examples --examples --release
 ```
 
-To build Linux-only modules, use the "linux" feature:
+To build Linux-only modules, use the `linux` feature:
 ```
 cargo build --package=examples --examples --features=linux --release
 ```
@@ -43,13 +49,13 @@ the SDK. You can start NGINX directly from this directory if you want to test th
 
 ### Mac OS dependencies
 
-In order to use the optional GNU make build process on MacOS, you will need to install additional tools. This can be
+To use the optional GNU build process on MacOS, you will need to install additional tools. This can be
 done via [homebrew](https://brew.sh/) with the following command:
 ```
 brew install make openssl grep
 ```
 
-Additionally, you may need to set up LLVM and clang. Typically, this is done as follows:
+Additionally, you may need to set up LLVM and Clang. Typically, this is done as follows:
 
 ```
 # make sure xcode tools are installed
@@ -64,16 +70,16 @@ See the [Dockerfile](Dockerfile) for dependencies as an example of required pack
 
 ### Build example
 
-Example modules are available in [examples](examples) folder. You can use `cargo build --package=examples --examples` to build these examples. After building, you can find the `.so` or `.dylib` in the `target/debug` folder. Add `--features=linux` to build linux specific modules. **NOTE**: adding the "linux" feature on MacOS will cause a build failure.
+Example modules are available in [examples](examples) folder. You can use `cargo build --package=examples --examples` to build these examples. After building, you can find the `.so` or `.dylib` in the `target/debug` folder. Add `--features=linux` to build Linux-specific modules. **NOTE**: adding the `linux` feature on MacOS will cause a build failure.
 
-For example (all examples plus linux specific):
+For example (all examples plus Linux specific):
 `cargo build --packages=examples --examples --features=linux`
 
 ### Docker
 
 We provide a multistage [Dockerfile](Dockerfile):
 
-    # build all dynamic modules examples and specify NGINX version to use
+    # Build all dynamic module examples and specify the NGINX version to use
     docker buildx build --build-arg NGX_VERSION=1.23.3 -t ngx-rust .
 
     # start NGINX using [curl](examples/curl.conf) module example:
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -14,23 +14,29 @@ dynamic modules completely in Rust.
 
 In short, this SDK allows writing NGINX modules using the Rust language.
 
+It contains the following Rust crates:
+ * [nginx-sys](./nginx-sys) - allows to use ngx_* C functions via FFI when implementing modules. The `-sys` is used to follow a [naming convention](https://doc.rust-lang.org/cargo/reference/build-scripts.html#-sys-packages) to link with a C library.
+ * [ngx](./src) - it an opinionated SDK to make it a bit easer to use [nginx-sys](./nginx-sys) crate. Implements `macro_rules`, provides a way to build a dynamic module without any C code (see `ngx_modules!` macro_rule).
```
f5yacobucci, line 19: ~~it~~ an opinionated SDK

README.md
```diff
@@ -14,23 +14,29 @@ dynamic modules completely in Rust.
...
 NGINX modules can be built against a particular version of NGINX. The following environment variables can be used to specify a particular version of NGINX or an NGINX dependency:
 
-* `ZLIB_VERSION` (default 1.2.13) -
+* `ZLIB_VERSION` (default 1.3)
 * `PCRE2_VERSION` (default 10.42)
 * `OPENSSL_VERSION` (default 3.0.7)
 * `NGX_VERSION` (default 1.23.3) - NGINX OSS version
 * `NGX_DEBUG` (default to false)-  if set to true, then will compile NGINX `--with-debug` option
+* `NGX_SRC_DIR` (default not set) - When the value is set, then use this NGINX source code folder to build bindings from
+* `NGX_CONFIGURE_ARGS` (default not set) - When the value is set, then run the NGINX configure script with
```
f5yacobucci, line 31: Suggest: When the value is set, use these arguments with the NGINX configure script

examples/README.md
```diff
@@ -28,7 +28,7 @@ cargo build --package=examples --examples
 
 This module demonstrates how to create a minimal dynamic module with `http_request_handler`, that checks for User-Agent headers and returns status code 403 if UA starts with `curl`, if a module is disabled then uses `core::Status::NGX_DECLINED` to indicate the operation is rejected, for example, because it is disabled in the configuration (`curl off`). Additionally, it demonstrates how to write a defective parser.
 
-An example of nginx configuration file that uses that module can be found at [curl.conf](./curl.conf).
+An example of an Nginx configuration file that uses that module can be found at [curl.conf](./curl.conf).
```
f5yacobucci, line 31: I haven't been checking these elsewhere, but I'm pretty sure the convention is to all-caps NGINX.

I'll confirm with Jodie.

examples/awssig.rs
```diff
@@ -334,11 +335,15 @@ http_request_handler!(awssigv4_header_handler, |request: &mut Request| {
...
-    for (name, value) in request.headers_out_iterator() {
-        ngx_log_debug_http!(request, "headers_out {}: {}", name, value);
+    for h in request.headers_out_iterator() {
+        ngx_log_debug_http!(request, "headers_out {}", h);
     }
-    for (name, value) in request.headers_in_iterator() {
-        ngx_log_debug_http!(request, "headers_in  {}: {}", name, value);
+    for mut h in request.headers_in_iterator() {
+        ngx_log_debug_http!(request, "headers_in  {}", h);
+        if h.lowercase_key() == Some("foo") {
```
f5yacobucci, line 343: Is this testing code? What's the 'foo' header in this context?

examples/curl.rs
```diff
@@ -4,9 +4,9 @@ use ngx::ffi::{
     ngx_uint_t, NGX_CONF_TAKE1, NGX_HTTP_LOC_CONF, NGX_HTTP_MODULE, NGX_RS_HTTP_LOC_CONF_OFFSET,
     NGX_RS_MODULE_SIGNATURE,
 };
-use ngx::http::MergeConfigError;
-use ngx::{core, core::Status, http, http::HTTPModule};
-use ngx::{http_request_handler, ngx_log_debug_http, ngx_modules, ngx_null_command, ngx_string};
+use ngx::{core, core::Status, http, http::HTTPModule, http::MergeConfigError};
+use ngx::{http_request_handler, ngx_log_debug, ngx_modules, ngx_null_command, ngx_string};
```
f5yacobucci, line 8: This is really just stylistic, but why not all in one:

`use ngx::{core, core::Status, http, http::{HTTPModule, MergeConfigError}, http_request_handler, ngx_log_debug, ngx_modules, ngx_null_command, ngx_string};`

examples/curl.rs
```diff
@@ -104,18 +104,30 @@ impl http::Merge for ModuleConfig {
 }
 
 http_request_handler!(curl_access_handler, |request: &mut http::Request| {
+    // we can check if a request is internal and disable handler
+    let log = request.log();
+    ngx_log_debug!(log, "is internal: {}", request.is_internal());
```
f5yacobucci, line 109: When this is printed will there be enough context for a user to know what's internal? Is there a convention around module identification for context?

nginx-sys/build.rs
```diff
@@ -89,14 +90,14 @@ const ENV_VARS_TRIGGERING_RECOMPILE: [&str; 9] = [
 /// Function invoked when `cargo build` is executed.
 /// This function will download NGINX and all supporting dependencies, verify their integrity,
 /// extract them, execute autoconf `configure` for NGINX, compile NGINX and finally install
-/// NGINX in a subdirectory with the project.
+/// NGINX in a subdirectory with the project or use provided NGNINX source path
```
f5yacobucci, line 93: NGNINX -> NGINX

nginx-sys/build.rs
```diff
@@ -430,57 +469,56 @@ fn extract_all_archives(cache_dir: &Path) -> Result<Vec<(String, PathBuf)>, Box<
...
+
+    // TODO: a problem: NGX_SRC_DIR was provided and user already ran configure/build
+    // but did not provide NGX_CONFIGURE_ARGS and the flags used are different with our default.
+    // what do to?
+    if !autoconf_makefile_exists {
+        run_configure_nginx(nginx_src_dir, &flags)?;
     }
-    Ok((nginx_install_dir, nginx_src_dir.to_owned()))
+
+    // TODO: a problem why we need to ran build - openssl needs to be configured if it is used,
```
f5yacobucci, line 504: Is this, or should this, be filed as a github issue?

src/core/string.rs
```diff
@@ -56,6 +57,15 @@ impl NgxStr {
         &self.0
     }
 
+    pub fn as_str(&self) -> &str {
+        // Safety: Converting the raw data to a string slice is unsafe, but as long
+        // as the lifetime of NgxStr is properly managed, this should be safe.
+        unsafe {
+            // let slice = std::slice::from_raw_parts(self.0.data, self.0.len);
+            std::str::from_utf8_unchecked(&self.0)
```
f5yacobucci, line 65: Can we be guaranteed this is valid UTF-8? to_str is using the checked version now.

src/http/headers.rs
```diff
@@ -0,0 +1,137 @@
...
+
+impl Header {
+    pub fn set_value(&mut self, value: &str) {
+        // we can use pool to allocate and then copy
+        unsafe { self.0.as_mut() }.map(|table| {
+            table.value.len = value.len() as _;
+            table.value.data = crate::ffi::str_to_uchar(self.1, value);
+        });
+
+        // Alternative way is using CString and transfer ownership.
```
f5yacobucci, line 20: Any speculation which one might be more efficient?

I'm wagering CString will put memory management under Rust semantics and reference counting. Where pool will allocate from nginx, could this mean it gets held longer in the pool?

src/http/headers.rs
```diff
@@ -0,0 +1,137 @@
...
+    pub fn set_key(&mut self, key: &str) {
+        // we can use pool to allocate and then copy
+        unsafe { self.0.as_mut() }.map(|table| {
+            table.key.len = key.len() as _;
+            table.key.data = crate::ffi::str_to_uchar(self.1, key);
+            table.lowcase_key = crate::ffi::str_to_uchar(self.1, key.to_lowercase().as_str());
+        });
+    }
+
+    pub fn set_hash(&mut self, hash: ngx_uint_t) {
```
f5yacobucci, line 38: Overall comment, we'll need doc strings for the API. I don't get what this is doing, is it setting the hash algorithm? Is hash a closed set of values, if so, an enum could be more expressive.

src/http/headers.rs
```diff
@@ -0,0 +1,137 @@
...
+        unsafe {
+            std::str::from_utf8(std::slice::from_raw_parts(
+                unsafe { *self.0 }.value.data,
+                unsafe { *self.0 }.value.len as usize,
+            ))
+            .unwrap()
+        }
+    }
+
+    pub fn key(&self) -> &str {
```
f5yacobucci, line 54: This looks like a collections (map) API. Is there a collection trait we can implement for a consistent Rust idiomatic API?

Does this follow NGINX patterns?

I think this should either be clearly recognizable as a Rust set of functions or an NGINX set of functions, my opinion leans more towards something Rust-like, especially if there's a trait.

ivanitskiy: there is no trait in Rust. there is `pub struct HeaderMap<T = HeaderValue> { /* private fields */ }` which is here. and this struct has some methods. see: 
https://docs.rs/http/latest/http/header/struct.HeaderMap.html#

Unfortunately, we are constrained with NGINX types and implementations.

so I'm following Nginx way. in Rust's  `Header` is an abstraction that encapsulates table_elt and pool: `pub struct Header(*mut ngx_table_elt_t, *mut ngx_pool_t);` and it represents a single Header (a pair of key and value). 
in nginx you don't remove headers, you set hash to `0` ( tl;dr - headers are lists of lists ).

the next thing is to have a HeaderMap (constructed from `headers_in.headers` or `headers_out.headers` which has a set of methods similar to hyperium's`http` crate. 

anyways, seems like what I've done here at all too. if you have some ideas please share

src/http/headers.rs
ivanitskiy: This is magic from the C guide:

from https://nginx.org/en/docs/dev/development_guide.html#list

> For example, to mark HTTP output headers (which are stored as ngx_table_elt_t objects) as missing, set the hash field in ngx_table_elt_t to zero. Items marked in this way are explicitly skipped when the headers are iterated over.

it is trick to "delete" header (by marking hash =0 for that element)

ivanitskiy: basically set_key, set_hash, set_value mimic the NGINX's API.

PR #59, opened 2024-02-20
fix: zlib update to 1.3.1

allan2:
The previous 1.3 link is dead. 1.3.1 was released on Jan 22, 2024.


ivanitskiy:
This seems will be quite an issue when we need to update code because of how zlib hosts artifacts. And requres to think how to avoid this kind of issue holistically. 

As a workaround to unblock you can follow
From
https://github.com/nginxinc/ngx-rust/issues/41#issuecomment-1692410297

Create/update `.cargo/config.toml` and add 

```toml
[env]
ZLIB_VERSION = "1.3"
```


Or simply setting env variable  and then compile `ZLIB_VERSION=1.2.13 cargo build -p examples --examples`

PR #66, opened 2024-03-06
fix: use std::ffi::c_char instead of platform specific int

dekobon:
### Proposed changes
This change reverts a previous change that did platform detection and set a type correspondingly to u8 or i8. By using std::ffi::c_char we get a portable way to do the same thing.

This PR fixes a shortcoming with the change in commit: d66f3418d64a4acf18d6b9139db89cd0ec775690

### Checklist

- [X] I have written my commit messages in the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format. 
- [X] I have read the [CONTRIBUTING](/CONTRIBUTING.md) doc
- [X] I have added tests (when possible) that prove my fix is effective or that my feature works
- [X] I have checked that all unit tests pass after adding my changes
- [X] I have updated necessary documentation
- [X] I have rebased my branch onto master
- [X] I will ensure my PR is targeting the main branch and pulling from my branch from my own fork

### DIFF
--- a/examples/upstream.rs
+++ b/examples/upstream.rs
@@ -250,7 +250,7 @@ unsafe extern "C" fn ngx_http_upstream_init_custom(
             NGX_LOG_EMERG as usize,
             cf,
             0,
-            "CUSTOM UPSTREAM no upstream srv_conf".as_bytes().as_ptr() as *const Ptr,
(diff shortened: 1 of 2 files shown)


dekobon:
@bavshin-f5 Thank you for the correction regarding the type usage.

PR #72, opened 2024-04-02
feat: modify cache & NGINX dir

JyJyJcr:
### Proposed changes
This PR add some environment variable options useful to develop modules using the SDK.

#### `CACHE_DIR`

Default: `[nginx-sys's root]/.cache`

Specify cache directory. With relative -> absolute path converting option provided in `.cargo/config.toml` like below, we are able to generate a project-owned cache dir:

```toml
[env]
CACHE_DIR = { value = ".cache", relative = true }
```

This option makes it easier to save and restore the cache, which speeds up CI/CD.

##### Sidenote

Originally the default location of the cache dir is in the parent dir of the manifest dir (= root dir) of `nginx-sys`: this is equivalent to set the default of `CACHE_DIR` as `[nginx-sys's root]/../.cache`. This setting resulted the cache dir to be generated in the root dir of `ngx-rust` as long as `ngx-rust` is treated as the main project (this is likely the originally intended behavior).
However, once `nginx-sys` is whether directly or indirectly used as a dependency of other projects, the repository was copied in `$CARGO_HOME/registry/cache/index.crates.io-***/nginx-sys-***/` or `$CARGO_HOME/git/checkouts/ngx-rust-***/***/`, causing the cache dir to be generated in the same depth with the copies of crates. This was apparently undesirable.
So the default value of `CACHE_DIR` is changed to `[nginx-sys's root]/.cache` to prevent the "pollution" described above. Instead, `ngx-rust` also use `CACHE_DIR` to move the cache dir to the original location.

#### `NGINX_INSTALL_ROOT_DIR`

Default: `{CACHE_DIR}/nginx`

Specify the directory where the NGINX compiled in the ***series*** of build will be installed. By default, each of Nginx is installed in the sub-subdirectory labeled by the version, OS and Architecture. This option is useful at developing stage to try multiple versions with your own conf files (for the latter aim only, `NGINX_INSTALL_DIR` is more suited).

#### `NGINX_INSTALL_DIR`

Default:  `{NGINX_INSTALL_ROOT_DIR}/{NGX_VERSION}/{OS}-{Arch}`

Specify the directory where the NGINX compiled in ***this*** build will be installed. This option is mainly intended to facilitate the test automation in CI/CD.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,9 +18,9 @@ In short, this SDK allows writing NGINX modules using the Rust language.
 
 NGINX modules can be built against a particular version of NGINX. The following environment variables can be used to specify a particular version of NGINX or an NGINX dependency:
 
-* `ZLIB_VERSION` (default 1.3.1) -
-* `PCRE2_VERSION` (default 10.42 for NGINX 1.22.0 and later, or 8.45 for earlier)
-* `OPENSSL_VERSION` (default 3.2.1 for NGINX 1.22.0 and later, or 1.1.1w for earlier)
+* `ZLIB_VERSION` (default 1.3.1) - zlib version
+* `PCRE2_VERSION` (default 10.42 for NGINX 1.22.0 and later, or 8.45 for earlier) - PCRE1 or PCRE2 version
+* `OPENSSL_VERSION` (default 3.2.1 for NGINX 1.22.0 and later, or 1.1.1w for earlier) - OpenSSL version
 * `NGX_VERSION` (default 1.24.0) - NGINX OSS version
 * `NGX_DEBUG` (default to false)-  if set to true, then will compile NGINX `--with-debug` option
 
@@ -38,8 +38,14 @@ cargo build --package=examples --examples --features=linux --release
 After compilation, the modules can be found in the path `target/release/examples/` ( with the `.so` file extension for
 Linux or `.dylib` for MacOS).
 
-Additionally, the folder  `.cache/nginx/{NGX_VERSION}/{OS}/` will contain the compiled version of NGINX used to build
+Additionally, the folder  `.cache/nginx/{NGX_VERSION}/{OS}-{Arch}/` will contain the compiled version of NGINX used to build
 the SDK. You can start NGINX directly from this directory if you want to test the module.
+
+The following environment variables can be used to change locations of cache directory and NGINX directory:
+
+* `CACHE_DIR` (default `[nginx-sys's root directory]/.cache`) - the directory containing cache files, means PGP keys, tarballs, PGP signatures, and unpacked source codes. It also contains compiled NGINX in default configuration.
+* `NGINX_INSTALL_ROOT_DIR` (default `{CACHE_DIR}/nginx`) - the directory containing the series of compiled NGINX in its subdirectories
+* `NGINX_INSTALL_DIR` (default `{NGINX_INSTALL_BASE_DIR}/{NGX_VERSION}/{OS}-{Arch}`) - the directory containing the NGINX compiled in the build
 
 ### Mac OS dependencies
 
--- a/README.md
+++ b/README.md
@@ -38,14 +38,14 @@ cargo build --package=examples --examples --features=linux --release
 After compilation, the modules can be found in the path `target/release/examples/` ( with the `.so` file extension for
 Linux or `.dylib` for MacOS).
 
-Additionally, the folder  `.cache/nginx/{NGX_VERSION}/{OS}-{Arch}/` will contain the compiled version of NGINX used to build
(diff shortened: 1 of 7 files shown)


dekobon:
Let's add this to the `.gitignore file`
```
nginx-sys/.cache/
```

README.md
```diff
@@ -38,9 +38,15 @@ cargo build --package=examples --examples --features=linux --release
 After compilation, the modules can be found in the path `target/release/examples/` ( with the `.so` file extension for
 Linux or `.dylib` for MacOS).
 
-Additionally, the folder  `.cache/nginx/{NGX_VERSION}/{OS}/` will contain the compiled version of NGINX used to build
+Additionally, the folder  `.cache/nginx/{NGX_VERSION}/{OS}-{Arch}/` will contain the compiled version of NGINX used to build
```
dekobon, line 41: If we are going to encode architecture here, we may want to consider over to using the full [cargo triple string](https://doc.rust-lang.org/cargo/appendix/glossary.html#target) because os and architecture are not enough information to tell you if you can actually run the binary on a system. For example, a library may be build for `musl`, and will not work on other glibc variants.

It looks like `<arch><sub>-<vendor>-<sys>-<abi>`. 

What do you think of doing that instead?

dekobon (changes requested):
This is a solid PR! Thank you.

Could you please add the new .cache directory to the .gitignore file?
Also, please reply to the comment about using the cargo triple instead of the os-arch pattern. I do not have a strong opinion, so I would love to know what you think.

Once again, thank you for the contribution.

README.md
JyJyJcr: Cargo triple strings contain enough information to replace `{OS}-{Arch}` since we currently only support linux and macos (if not, there are more variants listed [here](https://doc.rust-lang.org/std/env/consts/constant.OS.html) and [here](https://doc.rust-lang.org/std/env/consts/constant.ARCH.html)). What I suspect is whether the triple is enough. Does Nginx binary vary depending on Linux distributions? For example, do we have to distinguish two same version Nginxs, one of which is compiled on Debian and the other is on CentOS?

dekobon: > Does Nginx binary vary depending on Linux distributions

It does only for the version created by the distros for their packaging systems. Many of those contain patches. Those binaries are considered "unsupported" / "unsupportable" from the perspective of NGINX. However, that doesn't really matter for ngx-rust, because we are building the binaries from scratch without any such patches, so I believe we do not need to encode any distro specific information. In short, we don't need to care about this for ngx-rust.

Let's change the PR to support the triple string, and I'll click the merge button.

dekobon:
> Thank you for your rapid review.
> 
> > Let's add this to the `.gitignore file`
> > ```
> > nginx-sys/.cache/
> > ```
> 
> We don't have to add this, because `ngx-rust` set `CACHE_DIR` to `[ngx-rust's root]/.cache`:
> 
> https://github.com/nginxinc/ngx-rust/blob/29076150102b5524df635ab114bca1160f8e7989/.cargo/config.toml#L15-L16
> 
> `ngx-rust` keeps generating cache in its root dir, same as before.

Oh, that's my oversight. Thank you for clarifying.

JyJyJcr:
I changed the src dir name in the same manner to the Nginx dir. In addition, I added `target-triple` to dev-dependencies to fix `log_test.rs` to support new style of the directory structure.

README.md
dekobon, line 41: > `.cache/nginx/{NGX_VERSION}/{OS}-{Arch}/`
One last thing, let's update the string in the readme to reflect the triple format:

README.md
```diff
@@ -38,9 +38,15 @@ cargo build --package=examples --examples --features=linux --release
...
 the SDK. You can start NGINX directly from this directory if you want to test the module.
 
+The following environment variables can be used to change locations of cache directory and NGINX directory:
+
+* `CACHE_DIR` (default `[nginx-sys's root directory]/.cache`) - the directory containing cache files, means PGP keys, tarballs, PGP signatures, and unpacked source codes. It also contains compiled NGINX in default configuration.
+* `NGINX_INSTALL_ROOT_DIR` (default `{CACHE_DIR}/nginx`) - the directory containing the series of compiled NGINX in its subdirectories
+* `NGINX_INSTALL_DIR` (default `{NGINX_INSTALL_BASE_DIR}/{NGX_VERSION}/{OS}-{Arch}`) - the directory containing the NGINX compiled in the build
```
dekobon, line 48: Same here.

JyJyJcr:
Updated `README.md`. Thank you for the continuous support!

PR #73, opened 2024-04-06
fix: remove shared ref to static mut

JyJyJcr:
### Proposed changes

This PR removes shared references to mutable statics, which is announced to be treated as hard error in 2024 edition.
Link to the issue: https://github.com/rust-lang/rust/issues/114447
While the compiler suggests to use `addr_of!()` or `addr_of_mut!()`, In most cases we use `borrow()` because we need not raw pointers but references.

### Checklist


JyJyJcr:
Thank you for the review!

Now I replaced `borrow()` to `&*addr_of!()` following the recommended migration. I agree that using `borrow()` is defective.

I had tried to use `addr_of!()` to deal with the warning, but I couldn't find the proper way. Thank you for teaching me the pattern working correctly.

PR #74, opened 2024-04-07
feat: module of utilities for integration tests

JyJyJcr:
### Proposed changes

This PR adds `test_util` module, utilities for integration tests of Nginx modules. The utilities were moved from `tests/log_test.rs`, and fixed to use the bundled Nginx used in bindings, the install path of which is recorded in `nginx-sys::metadata::NGINX_INSTALL_DIR`.

### Checklist


dekobon:
Hi there, once again thank you for all of your contributions. I talked over this PR with some of the other project maintainers and there is a strong opinion that this functionality may be better in another project/library that supplements `ngx-rust`. Would it be possible to do that rather than integrate it into the core SDK?

JyJyJcr:
Of course possible (and I agree to split it from the core), but before changing let me clarify which is better, creating a new GitHub repository, or just separating it into a new subdirectory in this repository, just like `nginx-sys`?
Sidenote: we need to keep `nginx-sys::metadata` module to salvage build informations for integration tests in either case.

JyJyJcr:
> There's `nginx-sys::NGX_PREFIX` that already contains the installation path (NULL-terminated u8 slice, so you need to strip the last byte and use `std::os::unix::ffi::OsStrExt`).

Thank you for important information. Now `test_util` uses constants in `bindings.rs` (not only `NGX_PREFIX` but also other configurations, such as `NGX_CONF_PATH`).

JyJyJcr:
> Hi there, once again thank you for all of your contributions. I talked over this PR with some of the other project maintainers and there is a strong opinion that this functionality may be better in another project/library that supplements `ngx-rust`. Would it be possible to do that rather than integrate it into the core SDK?

Splitting the module to another project could cause incompatible (different version) two indirect dependencies to `ngx` crate. Such state would generate `binding.rs` twice, for two versions, resulting bad CI/CD performance.
Rather I add the feature `test_util`, and you can enable the module only in dev-dependencies.

PR #76, opened 2024-04-23
fix: silence remaining shared ref to static mut warning

bavshin-f5:
The example is disabled by default and was missed in #73


ivanitskiy:
lgtm

PR #83, opened 2024-06-11
feat(sys): update default NGINX version to 1.26.1

bavshin-f5:
1.24 branch is no longer supported, so switch to the latest stable release.


dekobon (approved):
LGTM!

PR #88, opened 2024-08-23
Resize vector holding ip address to correct size

jmccl:
Trivial change to resize the vector holding the ip address to the correct size. Without this the subsequent call to 'from_utf8' operates on a vector that is too large, resulting in garbage at the end of the string produced.


jmccl:
Good catch. I've updated the PR.

jmccl:
Done.

PR #90, opened 2024-09-03
Fix configuration lifetime and expose more config scopes

bavshin-f5:
Fix `get_module_loc_conf` result lifetime that made the Rust compiler think we're borrowing from the `Request`.
Expose `srv_conf` and `main_conf`.

### DIFF
--- a/src/http/request.rs
+++ b/src/http/request.rs
@@ -145,19 +145,49 @@ impl Request {
         unsafe { (*self.connection()).log }
     }
 
-    /// Module location configuration.
-    fn get_module_loc_conf_ptr(&self, module: &ngx_module_t) -> *mut c_void {
-        unsafe { *self.0.loc_conf.add(module.ctx_index) }
+    /// Global configuration for a module.
+    ///
+    /// Applies to the entire `http` block.
+    ///
+    /// # Safety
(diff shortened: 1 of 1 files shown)


src/http/request.rs
```diff
@@ -167,13 +194,9 @@ impl Request {
...
     /// Get Module context
     pub fn get_module_ctx<T>(&self, module: &ngx_module_t) -> Option<&T> {
-        let cf = self.get_module_ctx_ptr(module) as *mut T;
-
-        if cf.is_null() {
-            return None;
-        }
-        let co = unsafe { &*cf };
-        Some(co)
+        let ctx = self.get_module_ctx_ptr(module) as *const T;
```
xeioex, line 197: If we are using `.cast::<T>()` to cast to `*const T` above, shouldn't we use the same here?

xeioex (approved):
Otherwise looks good.

src/http/request.rs
bavshin-f5: `cast()` actually converts to `*mut T` and makes the debug code one (`mov`) instruction longer.
The release code looks the same though, so I'll make this change.

xeioex (approved):
Looks good.

PR #91, opened 2024-09-03
Byte-oriented methods for `ngx_str_t`

bavshin-f5:
`ngx_str_t` likely does *not* contain UTF-8, so converting it from/to UTF-8 strings is expensive and useless. `from_bytes()` and `as_bytes()` should be a more appropriate interface.

At some point in the future I intend to drop the `String`/`&str` conversion methods, as these aren't safe.

### DIFF
--- a/nginx-sys/src/lib.rs
+++ b/nginx-sys/src/lib.rs
@@ -75,19 +75,50 @@ pub unsafe fn str_to_uchar(pool: *mut ngx_pool_t, data: &str) -> *mut u_char {
 }
 
 impl ngx_str_t {
+    /// Returns the contents of this `ngx_str_t` as a byte slice.
+    ///
+    /// The returned slice will **not** contain the optional nul terminator that `ngx_str_t.data`
+    /// may have.
+    #[inline]
+    pub fn as_bytes(&self) -> &[u8] {
+        if self.is_empty() {
+            &[]
+        } else {
+            // SAFETY: `ngx_str_t` with non-zero len must contain a valid correctly aligned pointer
+            unsafe { slice::from_raw_parts(self.data, self.len) }
+        }
+    }
+
+    /// Returns the contents of this `ngx_str_t` as a mutable byte slice.
+    #[inline]
+    pub fn as_bytes_mut(&mut self) -> &mut [u8] {
+        if self.is_empty() {
+            &mut []
+        } else {
+            // SAFETY: `ngx_str_t` with non-zero len must contain a valid correctly aligned pointer
+            unsafe { slice::from_raw_parts_mut(self.data, self.len) }
+        }
+    }
+
+    /// Returns `true` if the string has a length of 0.
+    #[inline]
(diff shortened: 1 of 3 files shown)


nginx-sys/src/lib.rs
```diff
@@ -75,19 +75,50 @@ pub unsafe fn str_to_uchar(pool: *mut ngx_pool_t, data: &str) -> *mut u_char {
...
 
 impl ngx_str_t {
+    /// Returns the contents of this `ngx_str_t` as a byte slice.
+    ///
+    /// The returned slice will **not** contain the optional nul terminator that `ngx_str_t.data`
+    /// may have.
+    #[inline]
+    pub fn as_bytes(&self) -> &[u8] {
+        if self.is_empty() {
+            Default::default()
```
xeioex, line 100: Just curious, isn't `&[]` a more concise and explicit here? The same for the `as_bytes_mut()`.

nginx-sys/src/lib.rs
```diff
@@ -121,6 +136,15 @@ impl ngx_str_t {
         std::str::from_utf8(self.as_bytes()).unwrap()
     }
 
+    /// Create an `ngx_str_t` instance from a byte slice.
+    ///
+    /// # Safety
+    ///
+    /// The caller must provide a valid pointer to a memory pool.
+    pub unsafe fn from_bytes(pool: *mut ngx_pool_t, src: &[u8]) -> Option<Self> {
```
xeioex, line 144: `from_string()` and `from_str()`, both return Self, not `Option<Self>`, shouldn't we change them as well?

xeioex (approved):
Looks good.
Added a couple of small questions.

nginx-sys/src/lib.rs
```diff
@@ -69,7 +69,7 @@ pub use bindings::*;
 /// let ptr = str_to_uchar(pool, data);
 /// ```
 pub unsafe fn str_to_uchar(pool: *mut ngx_pool_t, data: &str) -> *mut u_char {
-    let ptr: *mut u_char = ngx_palloc(pool, data.len() as _) as _;
+    let ptr: *mut u_char = ngx_pnalloc(pool, data.len() as _) as _;
```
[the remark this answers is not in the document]
bavshin-f5: Thanks for pointing out.

Many, many years ago (before bindgen 0.61) [`bindgen::Builder::size_t_is_usize`](https://docs.rs/bindgen/latest/bindgen/struct.Builder.html#method.size_t_is_usize) defaulted to `false`. As we require newer version of bindgen, it's safe to assume that `size_t` is always converted to `usize` and remove all such casts. Done in a separate commit.

nginx-sys/src/lib.rs
bavshin-f5: Not sure if it's worth fixing the `str` conversion functions. I planned to remove these or at least mark as deprecated.

xeioex (approved):
Looks good.

PR #94, opened 2024-10-25
Logging improvements

### LINKED ISSUE #47: Provide more logging macros
ngx-rust doesn't provide a full set of convenience macros for logging.

We support specific debug logging, but for other logging the user still needs to call the C FFI functions themselves. Needing to provide unnecessary arguments, e.g., in the upstream.rs module example:

```
ngx_conf_log_error(
            NGX_LOG_EMERG as usize,
            cf,
            0,
            "CUSTOM UPSTREAM no upstream srv_conf".as_bytes().as_ptr() as *const i8,
       );
```

This can be simplified with a corresponding macro like we do for `ngx_log_debug!`.

### LINKED ISSUE #92: Crash in Error Handling in upstream.rs in examples
The four calls to [ngx_conf_log_error](https://github.com/nginxinc/ngx-rust/blob/3aedd7cf9742da927bef5f2c16596715936bccd7/examples/upstream.rs#L254) in upstream.rs coerce the error string to a '* const c_char', but the error string won't be null terminated so it will crash.

(I have not reproduced. I used the code as a template in my own module and it crashed. I fixed my own code via the discussion [here](https://users.rust-lang.org/t/converting-str-to-const-c-char/23115).)

bavshin-f5:
WIP, because I want to set up CI tests of the example modules before merging any significant changes.
***

Log macros now accept any object that implements `LogTarget`, i.e. requests, events, `ngx_conf_t`, etc. Default mask for the `ngx_log_debug!` is set according to the object type.
With all the complexity added, the code still makes 1-2 allocations less than the original variant and gets completely inlined in the release builds.

A few tricky moments:
 * `LogTarget` trait is intentionally kept [object-safe](https://doc.rust-lang.org/reference/items/traits.html#object-safety), because one of the planned further steps is to offer integration with the [`log`](https://docs.rs/log) facade and that required type erasure via `dyn LogTarget` (to be reevaluated).
 * Log methods take pointers to make things work uniformly for pointer and reference arguments. I wish I knew a better way to handle that.
 * Log methods take const ptr to avoid creating a mutable borrow (or failing to compile because the target was already borrowed immutably). It may look suspicious, but in the end we have a pointer to `ngx_log_t` that comes from C, points to a writable memory, and doesn't have to satisfy the Rust memory model constraints (as described in the [`std::ptr::from_ref`](https://doc.rust-lang.org/stable/std/ptr/fn.from_ref.html)).

Fixes #47
Fixes #92

### DIFF
--- a/examples/upstream.rs
+++ b/examples/upstream.rs
@@ -13,19 +13,21 @@ use std::slice;
 
 use ngx::core::{Pool, Status};
 use ngx::ffi::{
-    nginx_version, ngx_atoi, ngx_command_t, ngx_conf_log_error, ngx_conf_t, ngx_connection_t, ngx_event_free_peer_pt,
+    nginx_version, ngx_atoi, ngx_command_t, ngx_conf_t, ngx_connection_t, ngx_event_free_peer_pt,
     ngx_event_get_peer_pt, ngx_http_module_t, ngx_http_upstream_init_peer_pt, ngx_http_upstream_init_pt,
     ngx_http_upstream_init_round_robin, ngx_http_upstream_module, ngx_http_upstream_srv_conf_t, ngx_http_upstream_t,
     ngx_int_t, ngx_module_t, ngx_peer_connection_t, ngx_str_t, ngx_uint_t, NGX_CONF_NOARGS, NGX_CONF_TAKE1,
-    NGX_CONF_UNSET, NGX_ERROR, NGX_HTTP_MODULE, NGX_HTTP_UPS_CONF, NGX_LOG_EMERG, NGX_RS_HTTP_SRV_CONF_OFFSET,
+    NGX_CONF_UNSET, NGX_ERROR, NGX_HTTP_MODULE, NGX_HTTP_UPS_CONF, NGX_RS_HTTP_SRV_CONF_OFFSET,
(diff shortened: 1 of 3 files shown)


bavshin-f5:
After some research, I arrived to the following conclusions:
1. The following code is sound because `log: *mut ngx_log_t` has interior mutability, similar to `UnsafeCell`:
    ```rust
    impl LogTarget for ngx_event_t {
        fn get_log(&self) -> *mut ngx_log_t {
            self.log
        }
    }
    ```
2. The behavior of the following code is undefined, because `*mut ngx_log_t` obtained from a shared reference could be mutated in the `ngx_log_error_core`. Similar issue exists with `impl LogTarget for ngx_conf_t {}` and `ngx_conf_log_error`.
    ```rust
    impl LogTarget for ngx_log_t {
        fn get_log(&self) -> *mut ngx_log_t {
            ...
        }
    }
    ```
    Quoting the [language reference](https://doc.rust-lang.org/reference/behavior-considered-undefined.html#:~:text=Mutating%20immutable%20bytes),
    > the bytes pointed to by a shared reference, including transitively through other references (both shared and mutable) and Boxes, are immutable
3. All the cases of ` &something as *const _ as *mut _` we have in the code are potentially unsound and should be reviewed.

PR #95, opened 2024-10-29
test: integration tests for example modules

bavshin-f5:
* Test the whole process of building Rust code with `--add-dynamic-module` and the resulting modules with our Perl test framework
* Fix doctests for nginx-sys and add to CI
* Fix rustdoc warnings and add to CI
* Fix all remaining clippy warnings

### DIFF
--- a/.github/workflows/nginx.yaml
+++ b/.github/workflows/nginx.yaml
@@ -1,0 +1,96 @@
+name: NGINX
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+env:
+  CARGO_TERM_COLOR: 'always'
+  RUST_BACKTRACE: '1'
+  NGX_CONFIGURE: |
+    auto/configure
+      --with-compat
+      --with-debug
+      --with-http_realip_module
+      --with-http_ssl_module
+      --with-http_v2_module
+      --with-http_v3_module
+      --with-stream
+      --with-stream_realip_module
+      --with-stream_ssl_module
+      --with-threads
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        module:
+          # static
+          - dynamic
+        nginx-ref:
+          # master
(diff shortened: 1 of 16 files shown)


.github/workflows/nginx.yaml
```diff
@@ -0,0 +1,77 @@
...
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        nginx-ref:
+          # master
+          - stable-1.26
+
+    env:
```
thresheek, line 36: Those probably don't need to be defined for steps other than the actual prove run.

.github/workflows/nginx.yaml
```diff
@@ -0,0 +1,77 @@
...
+          restore-keys: ${{ runner.os }}-cargo-
+
+      - name: Configure nginx
+        working-directory: nginx
+        run: |
+          ${NGX_CONFIGURE} --add-dynamic-module=${{ github.workspace }}/examples
+
+      - name: Build nginx
+        working-directory: nginx
+        run: make
```
thresheek, line 74: Maybe -j something to speed things up

xeioex (commented):
Nitpicking: nginx_sys -> nginx-sys?

examples/t/upstream.t
```diff
@@ -0,0 +1,79 @@
...
+    upstream u {
+        server 127.0.0.1:8081;
+        custom 32;
+    }
+
+    server {
+        listen       127.0.0.1:8080;
+        server_name  localhost;
+
+        error_log %%TESTDIR%%/e_debug.log debug;
```
xeioex, line 46: Can't we test it without looking into debug log?

xeioex (approved):
Otherwise looks good.

examples/t/upstream.t
bavshin-f5: This module has no other testable behavior. And I wanted to see if the logs are working before I start breaking these.

PR #97, opened 2024-11-14
feat: conditional compilation support

bavshin-f5:
Nginx has various configuration options that may affect the availability of functions, constants and structure fields. This is not something that can be detected at runtime, as an attempt to use a symbol unavailable in the bindings would result in compilation error.

This PR introduces translation of a limited subset of `NGX_...` macros to the Rust `cfg` flags,  allowing conditional checks like `#[cfg(ngx_feature = "debug")]`.

There's no publicly available code that uses these flags _yet_, but I already need the functionality for some WIP stuff.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,16 +1,2 @@
-#
-# https://stackoverflow.com/questions/28124221/error-linking-with-cc-failed-exit-code-1
-#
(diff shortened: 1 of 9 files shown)


xeioex (approved):
Looks good.

PR #115, opened 2025-01-07
chore: update dependencies

bavshin-f5:
db065fa does not build with 1.79.0, but reordering the commits only makes things worse because `bindgen::RustTarget::from_str` in 0.70.x is unable to parse the full Rust version string.

In case if you are wondering why we are not using dependabot, here's the update task status:
<details>

<https://github.com/nginxinc/ngx-rust/network/updates/943285251>
```
Update check skipped
Dependabot updates have stopped for your repository due to repeated errors. To see logs, click "Check for Updates" and Dependabot will try again. The last error message was:
Dependabot can't resolve your Rust dependency files

Dependabot failed to update your dependencies because there was an error resolving your Rust dependency files.

Dependabot encountered the following error:

error: failed to load manifest for workspace member `dependabot_tmp_dir/examples`
referenced by workspace at `dependabot_tmp_dir/Cargo.toml`

Caused by:
  failed to parse manifest at `dependabot_tmp_dir/examples/Cargo.toml`

Caused by:
  the binary target name `examples` is forbidden, it conflicts with cargo's build directory names
```
</details>

I am aware of problem now and will eventually fix.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,18 +4,12 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.22.0"
+version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6e4503c46a5c0c7844e948c9a4d6acd9f50cccb4de1c48eb9e291ea17470c678"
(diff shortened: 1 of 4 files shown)


xeioex (approved):
LGTM.

Issue #107, opened 2024-12-24
Import optimization


bavshin-f5:
Dmitry has noted that the amount of imports required for even the smallest module example is overwhelming.

We can approach this from several directions:

1. Expose a minimal sufficient set of symbols via `ngx::core::prelude::*` and `ngx::http::prelude::*`
2. Use enums and `bitflags` as much as possible.
  E.g. wrapping `NGX_CONF_...` with `bitflags` would save a few imports and remove noisy `as ngx_uint_t` casts.
4. Reduce boilerplate code. For example, 
   ```rust
   #[no_mangle]
   #[used]
   pub static mut ngx_http_example_module: ngx_module_t = ngx_module_t {
     ctx: std::ptr::addr_of!(NGX_HTTP_EXAMPLE_MODULE_CTX) as _,
     // Safety: nginx does not modify module.commands.
     commands: NGX_HTTP_EXAMPLE_COMMANDS.as_ptr().cast_mut(),
     type_: NGX_HTTP_MODULE as _

     .. NGX_RS_MODULE_V1
   }
   ```
   would remove a couple of imports that are never necessary in the code and reduce amount of copy-paste needed to get started.

PR #117, opened 2025-01-09
chore: reduce boilerplate code for ngx_module_t declarations

xeioex:
This partially addresses #107.

### DIFF
--- a/examples/async.rs
+++ b/examples/async.rs
@@ -5,7 +5,7 @@ use std::time::Instant;
 
 use ngx::core;
 use ngx::core::prelude::*;
-use ngx::ffi::{ngx_array_push, ngx_http_core_run_phases, ngx_queue_s, NGX_RS_MODULE_SIGNATURE};
+use ngx::ffi::{ngx_array_push, ngx_http_core_run_phases, ngx_queue_s};
 use ngx::http::prelude::*;
 use ngx::http::{self, HTTPModule, MergeConfigError};
 use ngx::{http_request_handler, ngx_log_debug_http, ngx_null_command, ngx_string};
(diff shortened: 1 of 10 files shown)


examples/async.rs
```diff
@@ -25,8 +21,7 @@ impl http::HTTPModule for Module {
     unsafe extern "C" fn postconfiguration(cf: *mut ngx_conf_t) -> ngx_int_t {
         let cmcf = http::ngx_http_conf_get_module_main_conf(cf, &*addr_of!(ngx_http_core_module));
 
-        let h = ngx_array_push(&mut (*cmcf).phases[ngx_http_phases_NGX_HTTP_ACCESS_PHASE as usize].handlers)
-            as *mut ngx_http_handler_pt;
+        let h = ngx_array_push(&mut (*cmcf).phases[NGX_HTTP_ACCESS_PHASE].handlers) as *mut ngx_http_handler_pt;
```
bavshin-f5, line 24: I'd prefer a safer API for adding phases, restricted to `enum` instead of taking an arbitrary `usize`. In the meantime, it is not worth breaking the API for a cosmetic change.

bavshin-f5 (changes requested):
Still not convinced the commands initialization change is right. It assumes constness where we shouldn't, and could be less readable with a larger set of commands.

bavshin-f5 (changes requested):
`ngx_module_t::default` and `CONF_OFFSET` changes look good. I'm ready to cherry-pick these to master if you split the CONF_OFFSET patch and reorder both before the prelude one,

I have reservations about the prelude patch though. Maybe we need more high-level wrappers before introducing this.

Note: `refactor!:` prefix to the commit summaries seems more fitting. See https://www.conventionalcommits.org/en/v1.0.0/#summary and https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#-commit-message-guidelines

bavshin-f5:
Pushed https://github.com/nginx/ngx-rust/pull/117/commits/e8341496b87199669fa76e52525e97f1cd378bf2 and https://github.com/nginx/ngx-rust/pull/117/commits/4ee19c071ae3ef1759def71217b478601aa72fa3 to master with minor adjustments.

PR #132, opened 2025-02-13
refactor!: remove unnecessary declarative macros

bavshin-f5:
Remove a few overly direct translations of C macros to Rust declarative macros.

`ngx_string!` will be addressed separately, as the replacement has to ensure that the byte slice constants are NULL-terminated, and I haven't figured out the best way to enforce that (`fn from_static(src: &'static CStr) -> ngx_str_t` is the leading contender).


xeioex (commented):
```
There are no nginx core APIs that take a list of variables, terminated
by an empty entry. The variables are always added from the module code,
where we should use Rust iterators.
```

1) We do not need the comma in the first sentence.
2) The variables -> Variables (not specific variables).

```
There are no nginx core APIs that take a list of variables terminated
by an empty entry. Variables are always added from the module code,
where we should use Rust iterators.
```

xeioex (approved):
Otherwise looks good.

PR #151, opened 2025-04-16
 chore: reformat with max-width = 100 

bavshin-f5:
120 is too wide for diffing or editing in terminal.

### DIFF
--- a/nginx-sys/build/main.rs
+++ b/nginx-sys/build/main.rs
@@ -85,11 +85,17 @@ impl NginxSource {
         let source_dir = NginxSource::check_source_dir(source_dir).expect("source directory");
         let build_dir = NginxSource::check_build_dir(build_dir).expect("build directory");
 
-        Self { source_dir, build_dir }
+        Self {
+            source_dir,
+            build_dir,
+        }
     }
 
     pub fn from_env() -> Self {
-        match (env::var_os("NGINX_SOURCE_DIR"), env::var_os("NGINX_BUILD_DIR")) {
+        match (
+            env::var_os("NGINX_SOURCE_DIR"),
+            env::var_os("NGINX_BUILD_DIR"),
+        ) {
             (Some(source_dir), Some(build_dir)) => NginxSource::new(source_dir, build_dir),
             (Some(source_dir), None) => Self::from_source_dir(source_dir),
             (None, Some(build_dir)) => Self::from_build_dir(build_dir),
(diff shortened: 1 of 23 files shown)


nginx-sys/build/vendored.rs
```diff
@@ -204,7 +206,10 @@ fn all_archives() -> Vec<(String, String)> {
     let ngx_version = env::var("NGX_VERSION").unwrap_or_else(|_| NGX_DEFAULT_VERSION.into());
     let zlib_version = env::var("ZLIB_VERSION").unwrap_or_else(|_| ZLIB_DEFAULT_VERSION.into());
     // While Nginx 1.22.0 and later support pcre2 and openssl3, earlier ones only support pcre1 and openssl1. Here provides the appropriate (and as latest as possible) versions of these two dependencies as default, switching `***[major_version]_DEFAULT_VERSION` based on `is_after_1_22`. This facilitates to compile backport versions targeted for Nginx ealier than 1.22.0, which are still used in LTS releases of major Linux distributions.
```
xeioex, line 208: It seems the commentaries are not wrapped, may be we should do this manually? Here and in other places.

nginx-sys/build/main.rs
```diff
@@ -115,7 +125,10 @@ impl NginxSource {
...
         let source_dir = build_dir.parent().expect("source directory").to_path_buf();
 
-        Self { source_dir, build_dir }
+        Self {
+            source_dir,
+            build_dir,
+        }
     }
 
     #[cfg(not(feature = "vendored"))]
```
xeioex, line 134: Screenshot 2025-04-21 at 9 27 25 AM

Somehow this part of the file were not wrapped.

xeioex (changes requested):

nginx-sys/build/main.rs
bavshin-f5: rustfmt has [`format_strings`](https://rust-lang.github.io/rustfmt/?version=v1.8.0&search=#format_strings) option for splitting long strings, but it is unstable. Fixed manually.

nginx-sys/build/vendored.rs
bavshin-f5: Fixed with nightly `comment_width = 100` and `wrap_comments = true` options.

xeioex (approved):
Looks good.

PR #153, opened 2025-04-17
feat: raw ffi bindings to Stream module

bavshin-f5:


nginx-sys/build/wrapper.h
```diff
@@ -1,8 +1,23 @@
-#include <ngx_http.h>
-#include <ngx_conf_file.h>
 #include <ngx_config.h>
 #include <ngx_core.h>
 
+/* __has_include was a compiler-specific extension until C23,
+ * but it's safe to assume that bindgen supports it via libclang.
+ */
+#if defined(__has_include)
```
xeioex, line 7: Do we need to check for `defined(__has_include)` then?
and why `#include <ngx_stream.h>` is missing if __has_include is absent?

bavshin-f5: > Do we need to check for `defined(__has_include)` then?

I'd prefer to have this check, both because it is a recommended pattern of using `__has_include` and for symmetry with `expand_definitions()`.
And I'm sure that someone eventually will find the right combination of tools to fail this check. That I have no idea how does not mean it's not possible :)

>  and why `#include <ngx_stream.h>` is missing if __has_include is absent?

I thought it's pretty clear: we cannot verify that the stream support was enabled in this build of nginx and thus we don't want to expose stream symbols and allow accidentally linking to these. Even if it's a ~~dead code~~ fallback for a nearly impossible case, it should be correct.

PR #157, opened 2025-05-24
feat: a bit more useful NgxStr

bavshin-f5:
Prerequisite for the allocator-api work, but useful on its own.

* `nginx_sys::detail` could be another crate (`nginx-common`?), but there's just not enough code to split it now.
* no fancy `fmt::Debug` for `ngx_str_t`. Need to figure out how to override bindgen generated trait implementation without nuking Debug impls from most of the bindings. `Builder::no_debug()` does exactly that -- omits `Debug` on the `ngx_str_t` itself **and** on any structs that include `ngx_str_t` recursively.
* I attempted to implement precision and padding for `display_bytes`, but apparently precision requires a honest UTF-8 character width detection algorithm ([unicode-width](https://crates.io/crates/unicode-width)). The code without that can panic on an attempt to break an UTF-8 sequence. Maybe next time.
* Conversions and comparisons might be excessive. I took an inspiration from the [bstr](https://crates.io/crates/bstr) and left only types we realistically expect to use.
  I wish I could replace NgxStr with `bstr`, but it does not offer allocators support :(

### DIFF
--- a/nginx-sys/src/detail.rs
+++ b/nginx-sys/src/detail.rs
@@ -1,0 +1,112 @@
+//! Implementation details shared between nginx-sys and ngx.
+#![allow(missing_docs)]
+
+use core::fmt;
+
+#[inline]
+pub fn debug_bytes(f: &mut fmt::Formatter<'_>, bytes: &[u8]) -> fmt::Result {
+    if f.alternate() {
+        match bytes.len() {
+            0 => Ok(()),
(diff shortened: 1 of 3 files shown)


xeioex (approved):
> nginx_sys::detail could be another crate (nginx-common?), but there's just not enough code to split it now.

Agree.

Overall looks good to me.

Issue #85, opened 2024-06-19
crate documentation at docs.rs fails to build


bavshin-f5:
Neither of the crates has a documentation available. This happens because [docs.rs builds](https://docs.rs/about/builds) are performed in an isolated environment without network access and with limited filesystem access, therefore `nginx-sys` fails to fetch and build the dependencies.

<details>
<summary>Logs:</summary>

<https://docs.rs/crate/ngx/0.4.1/builds/1026822>
```
[...]

PR #160, opened 2025-05-28
feat!: add nginx-src crate with vendored nginx sources

bavshin-f5:
See #85 for rationale. Also "solves" all the issues with the dependency URLs, bad signatures etc.

Proof of concept. I'm planning to restore the download code and make it strictly opt-in before I mark this as ready for review.


bavshin-f5:
I think that's good enough to start looking. The only remaining item is README.md update.

nginx-src/src/download.rs
```diff
@@ -0,0 +1,257 @@
...
+        return Err(
+            format!("Downloaded file was not written to the expected location: {url}",).into(),
+        );
+    }
+    Ok(file_path)
+}
+
+/// Get a given tarball and signature file from a remote URL and copy it to the `.cache` directory.
+fn get_archive(cache_dir: &Path, source: &SourceSpec, version: &str) -> io::Result<PathBuf> {
+    let archive_url = (source.url)(version);
```
xeioex, line 163: Nitpicking: do we need the first pair of brackets here?

nginx-src/src/download.rs
```diff
@@ -0,0 +1,257 @@
...
+    let cache_dir = env::var("CACHE_DIR")
+        .map(PathBuf::from)
+        .unwrap_or(base_dir.join(".cache"));
+    if !cache_dir.exists() {
+        fs::create_dir_all(&cache_dir)?;
+    }
+    Ok(cache_dir)
+}
+
+/// Downloads a tarball from the specified URL into the `.cache` directory.
```
xeioex, line 136: Downloads, Get, Extract, Run, Returns   - grammar is a bit inconsistent in comments. 

I am fine both ways.  But it is better to be consistent.

xeioex (approved):
Over all looks good.

The source preparation code is more readable and maintainable now.

nginx-src/src/download.rs
bavshin-f5:
```
error[E0599]: no method named `url` found for reference `&SourceSpec<'_>` in the current scope
   --> nginx-src/src/download.rs:162:30
    |
162 |     let archive_url = source.url(version);
    |                              ^^^ field, not a method
    |
help: to call the function pointer stored in `url`, surround the field access with parentheses
    |
162 |     let archive_url = (source.url)(version);
    |                       +          +

For more information about this error, try `rustc --explain E0599`.
```

xeioex (approved):
LGTM.

PR #161, opened 2025-05-30
ci: test windows builds

bavshin-f5:
We haven't marked this platform as officially supported, but it's been mostly functional for a while. Some of my to-be-published changes modify or wrap platform-dependent code and may regress this state.

The execution time is  about 10 minutes, with most of that being spent on OpenSSL. This might be eventually improved by using [jom](https://github.com/qt-labs/jom) and patching the configure scripts to accept prebuilt dependencies on Windows.


.github/workflows/nginx.yaml
```diff
@@ -61,45 +98,134 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        nginx-ref:
+          - master
+        module:
+          - static
+
+    steps:
+      - uses: actions/checkout@v4
```
thresheek: Not really related to the PR, but we should use full sha ids in actions instead of the mutable tags

.github/workflows/nginx.yaml
```diff
@@ -61,45 +98,134 @@ jobs:
...
+          repository: 'nginx/nginx'
+          path: 'nginx'
+      - uses: actions/checkout@v4
+        with:
+          repository: 'nginx/nginx-tests'
+          path: 'nginx/tests'
+          sparse-checkout: |
+            lib
+
+      - uses: dtolnay/rust-toolchain@stable
```
thresheek: Same here, please use a full sha id instead of a tag.

bavshin-f5: Full sha for this action will unfortunately require regular manual updates. Do we really have to?

xeioex (approved):
LGTM.

PR #164, opened 2025-06-17
feat: allocator-api and shared memory support

bavshin-f5:
Not quite at the state I wanted to reach, but should be good to start the review.

Known issues and annoyances:
* `NgxString` size is 4 words (ptr, len, capacity, alloc). `std::string::String` is the same though, so this can be ignored for now.
* There's no non-deprecated `Hasher` impl in `::core`. We may need to add another dependency or implement our own Hasher.
* The `RwLock` is just a tiniest bit evil. It supposed to work anywhere where nginx locks do, but I have doubts...
   I considered `pthread_rwlock_t` with `pthread_rwlockattr_setpshared`, but there's that one platform without POSIX threads.
* Some remaining casual unsoundness, esp. around the `RbTree`. We may consider making pure Rust implementations of the data structures with fallible allocation support as a separate crate in the future.
* Windows failure is due to `postconfiguration` not being invoked. Not reproducible locally with up-to-date toolchain.
* Still need to figure how to wrap `ngx_shm_zone_t` safely.


src/allocator.rs
```diff
@@ -37,6 +37,17 @@ where
...
+/// Creates a `NonNull` that is dangling, but well-aligned for this alignment.
+///
+/// See also [::core::alloc::Layout::dangling()]
+#[inline(always)]
+pub(crate) const fn dangling_aligned<T>(align: usize) -> NonNull<T> {
+    unsafe {
+        let ptr = ptr::null_mut::<T>().byte_add(align);
+        NonNull::new_unchecked(ptr)
+    }
+}
```
fitzgen: Do the following debug asserts make sense to add?

```rust
debug_assert!(align.is_power_of_two());
debug_assert!(align >= mem::align_of::<T>());
```

(Assuming that this is intended to create over-aligned `T` pointers, if just the minimum alignment of `T` is required, then this whole function could be replaced with calls to `NonNull::<T>::dangling()`.)

src/core/pool.rs
```diff
@@ -1,13 +1,89 @@
...
+        // XXX: there should be no cleanup handlers for the allocations made using this API.
+        // Violating that could result in the following issues:
+        //  - use-after-free on large allocation
+        //  - multiple cleanup handlers attached to a dangling ptr (these are not unique)
+        if layout.size() > 0 // 0 is dangling ptr
+            && (layout.size() > self.as_ref().max || layout.align() > NGX_ALIGNMENT)
+        {
+            ngx_pfree(self.0.as_ptr(), ptr.as_ptr().cast());
+        }
+    }
```
fitzgen: Does `ngx_pool_t` do non-noop `free`s if done in LIFO order? If so, then it would also make sense for it to support in-place `realloc`s for the last allocation, and override the provided default methods for `Allocator::grow` and `Allocator::shrink` here as well.

Allowing the last allocation to grow in-place, even in bump allocators, is particularly nice when building `Vec`s of an unknown number of elements and which end up resizing and growing their underlying storage.

src/core/pool.rs
```diff
@@ -1,13 +1,89 @@
...
+/// Non-owning wrapper for an [`ngx_pool_t`] pointer, providing methods for working with memory pools.
 ///
 /// See <https://nginx.org/en/docs/dev/development_guide.html#pool>
-pub struct Pool(*mut ngx_pool_t);
+#[derive(Clone, Debug)]
+#[repr(transparent)]
+pub struct Pool(NonNull<ngx_pool_t>);
+
+unsafe impl Allocator for Pool {
+    fn allocate(&self, layout: Layout) -> Result<NonNull<[u8]>, AllocError> {
```
fitzgen, line 22: It seems like `ngx_pcalloc` exists, so it probably makes sense to implement `Allocator::allocate_zeroed` as well.

src/core/slab.rs
```diff
@@ -0,0 +1,150 @@
...
+        let shpool = self.0.as_ptr();
+        unsafe { ngx_shmtx_lock(ptr::addr_of_mut!((*shpool).mutex)) };
+        LockedSlabPool(self.0)
+    }
+}
+
+/// Wrapper for a locked [`ngx_slab_pool_t`] pointer.
+pub struct LockedSlabPool(NonNull<ngx_slab_pool_t>);
+
+unsafe impl Allocator for LockedSlabPool {
```
fitzgen, line 116: Similarly, it may make sense to provide non-default implementations of `Allocator::{allocate_zeroed,grow,shrink}` here as well, if the underlying `ngx_slab_pool_t` supports those operations (even if only for the last allocation).

fitzgen (commented):
Hi! @pchickey asked me to look at the `Allocator` trait bits because I have some experience implementing and using that trait.

The `dangling_aligned` helper could probably be simplified a little bit if it didn't have a type parameter, and always returned a `NonNull<u8>`. The only caller I saw (unless I missed something) uses `T = u8` and so if the type parameter was removed, then that also removes some of the questions around `T`'s alignment versus the alignment dynamically passed in as an argument, makes some `debug_assert!`s unnecessary, etc.

The comments about overriding the provided default `Allocator` methods are really opportunistic and are things that would be nice *if* the underlying capability is there. (And if the underlying capability isn't there right now, you might want to consider whether it is worth adding in the future. See the comment about building up `Vec`s.)

But yeah, overall, looks good to me!

(Leaving this review as a "comment" because I don't want to presume myself to be an nginx expert who can authoritatively approve or request changes to PRs or anything like that; just sharing my thoughts after a once-over. Hopefully they are helpful!)

src/allocator.rs
bavshin-f5: The function supposed to be a substitute for the unstable `Layout::dangling()` with over-alignment requirements. It is indeed unnecessarily generalized and can just take an existing `&Layout` to address these concerns.

Thanks for pointing this out.

src/core/pool.rs
bavshin-f5: Unfortunately, `ngx_pool_t` is not capable of that. It does not even track the necessary information.

The pool internals are exposed though, so there's a possibility for implementing this optimization in future. In a fragile and dangerous way :(

src/core/pool.rs
bavshin-f5: `ngx_pcalloc` is a combination of `ngx_palloc` and `memset`, the same as the default implementation of `Allocator::allocate_zeroed`. A correct implementation of `Allocator::allocate_zeroed` for `ngx_pool_t` would inherit all of the `Allocator::allocate` complexity, swap `ngx_palloc` with `ngx_pcalloc` and zeroize the data in the remaining branches.
Besides the obvious maintenance cost increase, I assumed that the default  `Allocator::allocate_zeroed` has more potential for inlining and optimizations.

src/core/slab.rs
bavshin-f5: Again, not something `ngx_slab_pool_t` is capable of doing.
It's possible to make `grow` noop if it fits into the same slot, but that requires reproducing a part of the slab pool internals in Rust.

src/core/pool.rs
```diff
@@ -1,13 +1,89 @@
...
+
+        // Verify the alignment of the result
+        debug_assert_eq!(ptr.align_offset(layout.align()), 0);
+
+        let ptr = NonNull::new(ptr.cast()).ok_or(AllocError)?;
+        Ok(NonNull::slice_from_raw_parts(ptr, layout.size()))
+    }
+
+    unsafe fn deallocate(&self, ptr: NonNull<u8>, layout: Layout) {
+        // ngx_pfree is noop for small allocations unless NGX_DEBUG_PCALLOC is set.
```
xeioex: Typo: NGX_DEBUG_PCALLOC  -> NGX_DEBUG_PALLOC

src/core/string.rs
```diff
@@ -239,6 +575,59 @@ mod tests {
...
+        let borrowed: &NgxStr = string.as_bytes().into();
+        let owned = NgxString::try_from_bytes_in(&string, Global).unwrap();
+
+        assert_eq!(string.as_bytes(), owned);
+        assert_eq!(ngx_string, owned);
+        assert_eq!(borrowed, owned);
+        assert_eq!(b"test", owned);
+        assert_eq!(owned, string.as_bytes());
+        assert_eq!(owned, borrowed);
+        assert_eq!(owned, b"test");
```
xeioex, line 601: Missing `assert_eq!(owned, ngx_string);` ?

src/core/string.rs
```diff
@@ -239,6 +575,59 @@ mod tests {
...
+
+        // String comparisons via Deref<Target = NgxStr>
+        assert_eq!(string, *owned);
+        assert_eq!(ngx_string, owned);
```
xeioex: Duplicate test with on the line 592, `assert_eq!(ngx_string, *owned);` ?

xeioex (changes requested):
Some minor issues I found up to "feat: owned byte string type with Allocator support"

examples/shared_dict.rs
```diff
@@ -0,0 +1,432 @@
...
+    let smcf = unsafe {
+        conf.cast::<SharedDictMainConfig>()
+            .as_mut()
+            .expect("shared dict main config")
+    };
+
+    let args =
+        unsafe { slice::from_raw_parts_mut((*cf.args).elts as *mut ngx_str_t, (*cf.args).nelts) };
+
+    let name: ngx_str_t = args[1];
```
pchickey, line 134: Please validate args len first

examples/shared_dict.rs
```diff
@@ -0,0 +1,432 @@
...
+
+        let _ = shared.write().remove(key);
+    } else {
+        let alloc = unsafe { SlabPool::from_shm_zone(&*smcf.shm_zone).expect("slab pool") };
+
+        let Ok(key) = NgxString::try_from_bytes_in(key.as_bytes(), alloc.clone()) else {
+            return;
+        };
+
+        let value = unsafe { slice::from_raw_parts(v.data, v.len() as usize) };
```
pchickey: This could use a comment that indicates its semantically equivalent to the try_from_bytes_in above

src/allocator.rs
```diff
@@ -0,0 +1,73 @@
+use ::core::mem;
+use ::core::ptr::{self, NonNull};
+
+pub use allocator_api2::alloc::*;
```
pchickey: For a pub use, I'd prefer for this glob to be expanded, so that any expansions to this interface can be tracked separately from upgrades of the allocator_api2 crate

src/allocator.rs
```diff
@@ -0,0 +1,73 @@
...
+        let ptr = ptr::null_mut::<u8>().byte_add(layout.align());
+        NonNull::new_unchecked(ptr)
+    }
+}
+
+#[cfg(feature = "alloc")]
+mod impls {
+    use super::*;
+
+    use super::boxed::Box;
```
pchickey:
```suggestion
    use allocator_api2::boxed::Box;
```

src/core/rbtree.rs
```diff
@@ -0,0 +1,358 @@
...
+    }
+}
+
+/// Raw iterator over the `ngx_rbtree_t` nodes.
+///
+/// This iterator remains valid after removing the current item from the tree.
+pub struct RawIter<'a> {
+    tree: NonNull<ngx_rbtree_t>,
+    node: *mut ngx_rbtree_node_t,
+    _ph: PhantomData<&'a ()>,
```
pchickey: I'd probably have named this field `_lifetime`

src/core/rbtree.rs
```diff
@@ -0,0 +1,358 @@
...
+            // empty tree
+            ptr::null_mut()
+        } else {
+            unsafe { ngx_rbtree_min(t.root, t.sentinel) }
+        };
+
+        Self {
+            tree,
+            node,
+            _ph: Default::default(),
```
pchickey: Arbitrary, but I think its more idiomatic to use `PhantomData` here instead of `Default::default()`

src/core/rbtree.rs
```diff
@@ -0,0 +1,358 @@
...
+/// This iterator remains valid after removing the current item from the tree.
+pub struct RawIter<'a> {
+    tree: NonNull<ngx_rbtree_t>,
+    node: *mut ngx_rbtree_node_t,
+    _ph: PhantomData<&'a ()>,
+}
+
+impl<'a> RawIter<'a> {
+    /// Creates an iterator for the `ngx_rbtree_t`.
+    pub fn new(tree: NonNull<ngx_rbtree_t>) -> Self {
```
pchickey: I'm not sure if the lifetime is actually providing any safety guarantees here since it isn't linked to the lifetime of the tree itself. Should this be an `unsafe fn new` because the user is still obligated to make sure the RawIter doesn't outlive the tree?

src/core/rbtree.rs
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    fn into_kv(self) -> (K, V) {
+        (self.key, self.value)
+    }
+}
+
+/// Raw iterator over the `ngx_rbtree_t` nodes.
+///
+/// This iterator remains valid after removing the current item from the tree.
```
pchickey: I'm not sure if this struct actually needs to be `pub` - it looks to me like its private to the implementation of all the other structs, but maybe I missed something. It looks like the two places this gets used its paired with a PhantomData<(K, V)> so maybe that aspect can be folded into the "Raw" iter?

The second line in the dod comment needs elaboration. What is the current item in the tree, the Item? What is the operation to remove an Item from the tree? Maybe the comment about validity to perform some specified remove operation actually belongs on the methods for `RBTree::iter` and `RBTree::iter_mut`, and on Iter and IterMut, if it matters to the consumers of those structs/methods.

src/core/rbtree.rs
```diff
@@ -0,0 +1,358 @@
...
+/// Wrapper over the `ngx_rbtree_t`.
+#[derive(Debug)]
+pub struct RbTree<K, V, A>
+where
+    A: Allocator,
+{
+    tree: ngx_rbtree_t,
+    sentinel: NonNull<ngx_rbtree_node_t>,
+    alloc: A,
+    _ph: PhantomData<(K, V)>,
```
pchickey: I'd probably name this field _kv_type to make it clear what the purpose of the PhantomData is

src/core/rbtree.rs
```diff
@@ -0,0 +1,358 @@
...
+}
+
+struct Node<K, V> {
```
pchickey: I may be misunderstanding exactly what this private helper is used for (it should have a comment) but if the offset_of the node field is important, should this be #[repr(C)] to make sure the offset of `node` matches the C repr? It looks like the Layout of this gets passed to the allocator, do we need to guarantee that it has the same layout as the C repr?

src/sync.rs
```diff
@@ -0,0 +1,123 @@
+//! Synchronization primitives over shared memory.
+//!
+//! This module provides an alternative implementation for the `ngx_atomic_t` type,
+//! `ngx_atomic_*`/`ngx_rwlock_*` family of functions and related usage patterns from nginx.
```
pchickey, line 4: This is a really excellent comment, thanks for writing it.

src/core/string.rs
```diff
@@ -196,8 +199,341 @@ impl_partial_ord_eq_from!(NgxStr, &'a [u8; N]; const N: usize);
...
+
     use super::*;
+
+    use crate::allocator::collections::TryReserveError;
+    use crate::allocator::vec::Vec;
+    use crate::allocator::{self, Allocator};
+
+    /// Owned byte string type with Allocator support.
+    ///
+    /// Inspired by `bstr` and `feature(bstr)`, with two important differences:
```
pchickey: Nit: link to bstr, such as https://doc.rust-lang.org/std/bstr/index.html or https://docs.rs/bstr/latest/bstr/

pchickey (commented):
Really nice work. Handful of pretty minor nits.

examples/shared_dict.rs
bavshin-f5: The len is validated by the config parser before invoking the callback. If it doesn't match the flags set in the `ngx_command_t`, it's totally fine to crash or fail in any other way.

I addressed this in a comment and added an assertion &mdash; likely redundant as Index ops are checked in debug.

src/core/rbtree.rs
bavshin-f5: After this comment I realized that I had no clear picture when writing `ngx::core::rbtree`, and thus failed in documenting the purpose.

The underlying tree is not quite a complete key-value map implementation. Sure, there is a key: `ngx_rbtree_key_t` (a.k.a. `usize`), and a value type which embeds the `ngx_rbtree_node_t`, but `RbTree<K,V,A>` type is not a safe wrapper for an arbitrary NGINX rbtree. It is an implementation of a `Map<K,V>` with `ngx_rbtree_t` as a building block.
`RawIter` though works on any `ngx_rbtree_key_t` and just accesses the tree nodes sorted by `node.key`, not being aware of the value type or allocation method.

In the end, I want to implement both a generic `ngx_rbtree_t` wrapper to work with references to the NGINX internal trees, and a high-level Map/Set/etc type for a pure Rust module code. I renamed things and added some comments to clarify that, but I don't believe I got the abstractions right yet.
I'll take some time to think about that while attending the conference.

The current implementation should be good enough for our internal projects though.

src/core/rbtree.rs
bavshin-f5: This type has no C counterpart and will not be accessed from C. AFAIU, the field offsets are stable when used within the same Rust compilation target, notwithstanding the layout or repr, so we don't need to force a specific representation.

bavshin-f5:
Appreciate the very detailed reviews.
Addressed most of the small nits with the last push, will take a pause to figure out how to improve the RbTree interface.

src/core/slab.rs
```diff
@@ -0,0 +1,149 @@
...
+        unsafe { self.0.as_mut() }
+    }
+}
+
+impl SlabPool {
+    /// Creates a new `SlabPool` from an initialized shared zone.
+    ///
+    /// # Safety
+    ///
+    /// Shared zone should be initialized and
```
xeioex: "Shared zone should be initialized and" 

"and" typo?

examples/t/shared_dict.t
```diff
@@ -0,0 +1,139 @@
...
+###############################################################################
+
+sub check {
+	my ($uri, $like) = @_; 
+
+	my $r = http_get($uri);
+
+	return unless ($r =~ $like && $r =~ /X-Process: (\d+)/);
+
+	return 1 if $^O eq 'MSWin32'; # only one active worker process
```
xeioex, line 107: Maybe we should check globally for 'MSWin32' and skip tests completely.

src/core/pool.rs
bavshin-f5: Apparently we already use in-place realloc in `ngx_array_push`, even if there's no general-purpose API for that.
I filed #169 to reimplement the optimization in our Allocator implementation.

PR #167, opened 2025-06-23
nginx-sys: define ngx_list_init

pchickey:
Like many of the functions defined in nginx-sys, `ngx_list_init` is defined in nginx as an inline function in a C header, so it isnt captured by bindgen. This definition is equivalent to the one in the C header: https://github.com/nginx/nginx/blob/master/src/core/ngx_list.h#L36-L52

### Proposed changes
Describe the use case and detail of the change. If this PR addresses an issue 
on GitHub, make sure to include a link to that issue here in this description 
(not in the title of the PR).

### Checklist

### DIFF
--- a/nginx-sys/src/lib.rs
+++ b/nginx-sys/src/lib.rs
@@ -183,6 +183,34 @@ pub fn ngx_timeofday() -> &'static ngx_time_t {
     unsafe { &*ngx_cached_time }
 }
 
+/// Initialize a list, using a pool for the backing memory, with capacity to store the given number
+/// of elements and element size.
+///
+/// # Safety
+/// * `list` must be non-null
+/// * `pool` must be a valid pool
+#[inline]
+pub unsafe fn ngx_list_init(
+    list: *mut ngx_list_t,
+    pool: *mut ngx_pool_t,
+    n: ngx_uint_t,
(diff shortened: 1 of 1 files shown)


nginx-sys/src/lib.rs
```diff
@@ -183,6 +183,34 @@ pub fn ngx_timeofday() -> &'static ngx_time_t {
...
+    size: usize,
+) -> ngx_int_t {
+    unsafe {
+        (*list).part.elts = ngx_palloc(pool, n * size);
+        if (*list).part.elts.is_null() {
+            return NGX_ERROR as ngx_int_t;
+        }
+        (*list).part.nelts = 0;
+        (*list).part.next = core::ptr::null_mut();
+        (*list).last = core::ptr::from_mut(&mut (*list).part);
```
bavshin-f5: `core::ptr::addr_of_mut!((*list).part)`?
Also, `core::ptr` is already in scope for this file. Can be shortened to `ptr::addr_of_mut...` to match the style we're trying to follow.

pchickey: Thanks!

bavshin-f5 (approved):
Thanks!

PR #170, opened 2025-07-01
feat: minimal async runtime on top of the NGINX event loop

bavshin-f5:
This change introduces a general infrastructure for spawing async tasks on NGINX event loop.  The only utility offered for now is a timer support via `ngx::async_::sleep`, with async IO and other scenarios being planned for future.

An incomplete connection IO example is available here: https://github.com/bavshin-f5/ngx-rust/tree/async

### DIFF
--- a/src/async_/sleep.rs
+++ b/src/async_/sleep.rs
@@ -1,0 +1,138 @@
+use core::future::Future;
+use core::mem;
+use core::pin::Pin;
+use core::ptr::{self, NonNull};
+use core::task::{self, Poll};
+use core::time::Duration;
+
+use nginx_sys::{ngx_add_timer, ngx_del_timer, ngx_event_t, ngx_log_t, ngx_msec_int_t, ngx_msec_t};
+use pin_project_lite::pin_project;
+
+use crate::{ngx_container_of, ngx_log_debug};
+
+/// Maximum duration that can be achieved using [ngx_add_timer].
+const NGX_TIMER_DURATION_MAX: Duration = Duration::from_millis(ngx_msec_int_t::MAX as _);
+
+/// Puts the current task to sleep for at least the specified amount of time.
+///
+/// The function is a shorthand for [Sleep::new] using the global logger for debug output.
+#[inline]
+pub fn sleep(duration: Duration) -> Sleep {
+    Sleep::new(duration, crate::log::ngx_cycle_log())
+}
+
+pin_project! {
+/// Future returned by [sleep].
+pub struct Sleep {
+    #[pin]
+    timer: TimerEvent,
+    duration: Duration,
+}
+}
+
+impl Sleep {
+    /// Creates a new Sleep with the specified duration and logger for debug messages.
+    pub fn new(duration: Duration, log: NonNull<ngx_log_t>) -> Self {
+        let timer = TimerEvent::new(log);
+        ngx_log_debug!(timer.event.log, "async: sleep for {duration:?}");
+        Sleep { timer, duration }
+    }
+}
+
+impl Future for Sleep {
+    type Output = ();
+
+    #[cfg(not(target_pointer_width = "32"))]
+    fn poll(self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> Poll<Self::Output> {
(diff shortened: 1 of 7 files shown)


src/async_/sleep.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    pub fn poll_sleep(
+        mut self: Pin<&mut Self>,
+        duration: ngx_msec_t,
+        context: &mut task::Context<'_>,
+    ) -> Poll<()> {
+        if self.event.timedout() != 0 {
+            Poll::Ready(())
+        } else if self.event.timer_set() != 0 {
+            self.waker = Some(context.waker().clone());
```
pchickey: Can we change this to use https://doc.rust-lang.org/beta/std/task/struct.Waker.html#method.clone_from, which may be substantially more efficient?

src/async_/spawn.rs
```diff
@@ -0,0 +1,147 @@
...
+        // process mode. Update `log` every time to avoid using stale log pointer.
+        self.event.log = unsafe { (*ngx_cycle).log };
+
+        // While this event is not used as a timer at the moment, we still want to ensure that it is
+        // compatible with `ngx_event_ident`.
+        if self.event.data.is_null() {
+            self.event.data = ptr::from_mut(self).cast();
+        }
+
+        // TODO: handle allocation failures
```
pchickey: VecDeque doesn't provide any mechanism to handle allocation failure, it will panic - and I think thats fine for this case. There's no recovering from the process heap getting exhausted

pchickey (approved):
Looks excellent. Thank you for iterating on this!

src/async_/spawn.rs
bavshin-f5: Generally, we treat any allocation in NGINX as fallible. We are expected to work in the environments with limited heap such as restricted cgroups or systems with overcommit disabled. It's better to drop a few connections in a predictable manner rather than lose a whole worker to the OOM killer.

It is my long-term goal to keep the Rust bindings to the same standards, even if we have to fight with the language sometimes.

pchickey:
```suggestion
        // FIXME: VecDeque::push could panic on an allocation failure, switch to a datastructure which will not and propagate the failure.
```

pchickey: Fair enough, I'm fine with working on that problem in the future, and I suggested a more specific comment which makes it clear what the intent is

bavshin-f5: Applied, thanks!

src/async_/spawn.rs
```diff
@@ -0,0 +1,149 @@
...
+    }
+}
+
+/// Creates a new task running on the NGINX event loop.
+pub fn spawn<F, T>(future: F) -> Task<T>
+where
+    F: Future<Output = T> + 'static,
+    T: 'static,
+{
+    ngx_log_debug!(unsafe { (*ngx_cycle).log }, "async: spawned new task");
```
xeioex: Missed "ngx_cycle_log().as_ptr()" replacement?

xeioex: "async: spawning new task"?

or move the debug below `runnable.schedule();`

src/async_/sleep.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    #[cfg(target_pointer_width = "32")]
+    fn poll(self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> Poll<Self::Output> {
+        if self.duration.is_zero() {
+            return Poll::Ready(());
+        }
+        let step = self.duration.min(NGX_TIMER_DURATION_MAX);
+
+        let mut this = self.project();
+        // Handle ngx_msec_t overflow on 32-bit platforms.
```
xeioex, line 60: It is hard to imagine ~49day timers in the wild, but lets keep it:)

xeioex (approved):
Look good.

src/async_/sleep.rs
bavshin-f5: ~25 days actually. Any more and our signed comparison magic considers the timer immediately expired.
And we just happen to have a module that needs to wake up every 60 days :)

PR #171, opened 2025-07-09
Refactored allocator-api2 reexports

bavshin-f5:
See individual commits.

`unsize_box` is a temporary solution, I'll revert and properly reexport it once I have time to send a patch to the upstream project.


xeioex (approved):
Looks good.

PR #174, opened 2025-07-15
fix: use CFLAGS from Makefile

### LINKED ISSUE #118: Use `CFLAGS` from the Makefile
Some compiler flags can affect the bindings or the binary in significant for us ways. At the very least, we should be able to find the preprocessor flags (`-D`) and pass to bindgen. That requires improving our Makefile parser and collecting more variables.

A practical example: I've been trying to enforce the alignment requirements by statically asserting that the `NGX_ALIGNMENT` the nginx binary was built with is large enough (improperly aligned pointers are UB in Rust, and some of the `core::` APIs verify alignment), and found that it wasn't possible to detect that the alignment was overridden via `auto/configure --with-cc-opt="-DNGX_ALIGNMENT=8"`.

ensh63:
### Proposed changes
Reads C include directories and compiler definitions from Makefile using ALL_INCS and CFLAGS variables.
The lists of include directories and definitions are used to compile `expand.c`.

Fixes #118 .


nginx-sys/build/main.rs
```diff
@@ -243,20 +244,27 @@ fn generate_binding(nginx: &NginxSource) {
...
+    nginx_autoconf_makefile_path: &PathBuf,
+) -> (Vec<PathBuf>, Vec<(String, Option<String>)>) {
+    fn add_includes(includes: &mut Vec<String>, line: &str) {
+        let mut words = line.split_ascii_whitespace();
+        while let Some(word) = words.next() {
+            if word == "-I" {
+                if let Some(inc) = words.next() {
+                    includes.push(inc.to_string());
+                }
+            } else if let Some(inc) = word.strip_prefix("-I") {
```
bavshin-f5: `"-I".strip_prefix("-I")` is `Some("")`.
Knowing this, and considering that `-I` without an argument is not a valid flag, we can simplify this block to 2 checks:
```rust
if let Some(value) = word.strip_prefix("-I") {
    let value = if value.is_empty() {
        words.next().expect("-I argument")
    } else {
        value
    };
    includes.push(value.to_string());
}
```

nginx-sys/build/main.rs
```diff
@@ -268,35 +276,47 @@ fn parse_includes_from_makefile(nginx_autoconf_makefile_path: &PathBuf) -> Vec<P
...
+        line += l;
+
+        if let Some(tail) = line.strip_prefix("ALL_INCS") {
+            add_includes(&mut includes, tail);
+        } else if let Some(tail) = line.strip_prefix("CFLAGS") {
+            add_includes(&mut cflags_includes, tail);
+
+            let words = line.split_ascii_whitespace();
+            for word in words {
+                if let Some(def) = word.strip_prefix("-D") {
```
bavshin-f5: `-D name=value` is a valid syntax for this parameter.

nginx-sys/build/main.rs
```diff
@@ -243,20 +244,27 @@ fn generate_binding(nginx: &NginxSource) {
...
-    fn extract_after_i_flag(line: &str) -> Option<&str> {
-        let mut parts = line.split("-I ");
-        match parts.next() {
-            Some(_) => parts.next().map(extract_include_part),
-            None => None,
+pub fn parse_makefile(
+    nginx_autoconf_makefile_path: &PathBuf,
+) -> (Vec<PathBuf>, Vec<(String, Option<String>)>) {
+    fn add_includes(includes: &mut Vec<String>, line: &str) {
+        let mut words = line.split_ascii_whitespace();
```
bavshin-f5: Do we need a shell-style word splitting here? If we do, [`shlex`](https://docs.rs/shlex/) is already in the dependency tree.

nginx-sys/build/main.rs
```diff
@@ -268,35 +276,47 @@ fn parse_includes_from_makefile(nginx_autoconf_makefile_path: &PathBuf) -> Vec<P
...
-                break;
+        line += l;
+
+        if let Some(tail) = line.strip_prefix("ALL_INCS") {
+            add_includes(&mut includes, tail);
+        } else if let Some(tail) = line.strip_prefix("CFLAGS") {
+            add_includes(&mut cflags_includes, tail);
+
+            let words = line.split_ascii_whitespace();
+            for word in words {
```
bavshin-f5: Have you considered making this a single pass? E.g., rename add_includes to parse_flags and collect both types of the parameters.
It could make the code a bit more readable, IMO.

bavshin-f5 (commented):
Please, pass the defines to bindgen via clang_args.

bavshin-f5 (approved):
Looks good to me, thanks!

PR #111, opened 2024-12-26
add `std`,`alloc` feature flag to support `no_std` build

JyJyJcr:
### Proposed changes
add features `std`, `alloc` to prepare for `no_std` build support.
- `alloc` feature means memory allocation used. Implementation differs depending on `std` feature:
  - with `std`: use `std` crate components. (e.g. `std::string::String`)
  - without `std`: use `alloc` crate components. (e.g. `alloc::string::String`)
- `std` feature means `std` crate components used. Currently the only difference between `std` and `alloc` is impl of `std::error::Error` (`core::error::Error` is not in Rust 1.79.0), which is not used.

`alloc` is the transitional feature: while it enables `no_std` build, it allows memory allocation using `std::alloc::System` internally. We can gradually replace `alloc` crate components by nginx-based ones.

For these features, many `std` imports are replaced by `core` ones.

Related issue (enhancement): https://github.com/nginxinc/ngx-rust/issues/109

### Checklist
Before creating a PR, run through this checklist and mark each as complete.

- [x] I have written my commit messages in the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format. 
- [x] I have read the [CONTRIBUTING](/CONTRIBUTING.md) doc
- [x] I have added tests (when possible) that prove my fix is effective or that my feature works
- [x] I have checked that all unit tests pass after adding my changes
- [x] I have updated necessary documentation
- [x] I have rebased my branch onto master
- [x] I will ensure my PR is targeting the main branch and pulling from my branch from my own fork

PR #197, opened 2025-08-18
add ngx::core::resolver, which provides async name resolution

pchickey:
Added behind its own feature flag, because it incurs two extra deps (thiserror and futures-channel) on top of the async feature set.

This PR is an upstreaming of nginx-acme::net::resolver, see: https://github.com/nginx/nginx-acme/blob/main/src/net/resolver.rs

There are no examples or tests for this module at this time.

### Checklist

### DIFF
--- a/src/async_/resolver.rs
+++ b/src/async_/resolver.rs
@@ -1,0 +1,336 @@
+// Copyright (c) F5, Inc.
+//
+// This source code is licensed under the Apache License, Version 2.0 license found in the
+// LICENSE file in the root directory of this source tree.
+
+//! Wrapper for the nginx resolver.
+//!
+//! See <https://nginx.org/en/docs/http/ngx_http_core_module.html#resolver>.
+
+use alloc::string::{String, ToString};
+use core::ffi::c_void;
+use core::fmt;
+use core::num::NonZero;
+use core::pin::Pin;
+use core::ptr::NonNull;
+use core::task::{Context, Poll, Waker};
+
+use crate::{
+    allocator::Box,
+    collections::Vec,
+    core::Pool,
+    ffi::{
+        ngx_addr_t, ngx_msec_t, ngx_resolve_name, ngx_resolve_start, ngx_resolver_ctx_t,
+        ngx_resolver_t, ngx_str_t,
+    },
+};
+use nginx_sys::{
+    NGX_RESOLVE_FORMERR, NGX_RESOLVE_NOTIMP, NGX_RESOLVE_NXDOMAIN, NGX_RESOLVE_REFUSED,
+    NGX_RESOLVE_SERVFAIL, NGX_RESOLVE_TIMEDOUT,
+};
+
+/// Error type for all uses of `Resolver`.
+#[derive(Debug)]
+pub enum Error {
+    /// No resolver configured
+    NoResolver,
+    /// Resolver error, with context of name being resolved
+    Resolver(ResolverError, String),
+    /// Allocation failed
+    AllocationFailed,
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Error::NoResolver => write!(f, "No resolver configured"),
+            Error::Resolver(err, context) => write!(f, "{err}: resolving `{context}`"),
+            Error::AllocationFailed => write!(f, "Allocation failed"),
+        }
+    }
+}
+impl core::error::Error for Error {}
+
+/// These cases directly reflect the NGX_RESOLVE_ error codes,
+/// plus a timeout, and a case for an unknown error where a known
+/// NGX_RESOLVE_ should be.
+#[derive(Debug)]
+pub enum ResolverError {
+    /// Format error (NGX_RESOLVE_FORMERR)
+    FormErr,
+    /// Server failure (NGX_RESOLVE_SERVFAIL)
+    ServFail,
+    /// Host not found (NGX_RESOLVE_NXDOMAIN)
+    NXDomain,
+    /// Unimplemented (NGX_RESOLVE_NOTIMP)
+    NotImp,
+    /// Operation refused (NGX_RESOLVE_REFUSED)
+    Refused,
+    /// Timed out (NGX_RESOLVE_TIMEDOUT)
+    TimedOut,
+    /// Unknown NGX_RESOLVE error
+    Unknown(isize),
+}
+impl fmt::Display for ResolverError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            ResolverError::FormErr => write!(f, "Format error"),
+            ResolverError::ServFail => write!(f, "Server Failure"),
+            ResolverError::NXDomain => write!(f, "Host not found"),
+            ResolverError::NotImp => write!(f, "Unimplemented"),
+            ResolverError::Refused => write!(f, "Refused"),
+            ResolverError::TimedOut => write!(f, "Timed out"),
+            ResolverError::Unknown(code) => write!(f, "Unknown NGX_RESOLVE error {code}"),
+        }
+    }
+}
+impl core::error::Error for ResolverError {}
+
+/// Convert from the NGX_RESOLVE_ error codes.
+impl From<NonZero<isize>> for ResolverError {
+    fn from(code: NonZero<isize>) -> ResolverError {
+        match code.get() as u32 {
+            NGX_RESOLVE_FORMERR => ResolverError::FormErr,
+            NGX_RESOLVE_SERVFAIL => ResolverError::ServFail,
+            NGX_RESOLVE_NXDOMAIN => ResolverError::NXDomain,
+            NGX_RESOLVE_NOTIMP => ResolverError::NotImp,
+            NGX_RESOLVE_REFUSED => ResolverError::Refused,
+            NGX_RESOLVE_TIMEDOUT => ResolverError::TimedOut,
+            _ => ResolverError::Unknown(code.get()),
+        }
+    }
+}
+
+type Res = Result<Vec<ngx_addr_t, Pool>, Error>;
+
+/// A wrapper for an ngx_resolver_t which provides an async Rust API
+pub struct Resolver {
+    resolver: NonNull<ngx_resolver_t>,
+    timeout: ngx_msec_t,
+}
+
+impl Resolver {
+    /// Create a new `Resolver` from existing pointer to `ngx_resolver_t` and
+    /// timeout.
+    pub fn from_resolver(resolver: NonNull<ngx_resolver_t>, timeout: ngx_msec_t) -> Self {
+        Self { resolver, timeout }
+    }
+
+    /// Resolve a name into a set of addresses.
+    pub async fn resolve_name(&self, name: &ngx_str_t, pool: &Pool) -> Res {
+        let mut resolver = Resolution::new(name, &ngx_str_t::empty(), self, pool)?;
+        resolver.as_mut().await
+    }
+
+    /// Resolve a service into a set of addresses.
+    pub async fn resolve_service(&self, name: &ngx_str_t, service: &ngx_str_t, pool: &Pool) -> Res {
+        let mut resolver = Resolution::new(name, service, self, pool)?;
+        resolver.as_mut().await
+    }
+}
+
+struct Resolution<'a> {
+    // Storage for the result of the resolution `Res`. Populated by the
+    // callback handler, and taken by the Future::poll impl.
+    complete: Option<Res>,
+    // Storage for a pending Waker. Populated by the Future::poll impl,
+    // and taken by the callback handler.
+    waker: Option<Waker>,
+    // Pool used for allocating `Vec<ngx_addr_t>` contents in `Res`. Read by
+    // the callback handler.
+    pool: &'a Pool,
+    // Owned pointer to the ngx_resolver_ctx_t.
+    ctx: Option<ResolverCtx>,
+}
+
+impl<'a> Resolution<'a> {
+    pub fn new(
+        name: &ngx_str_t,
+        service: &ngx_str_t,
+        resolver: &Resolver,
+        pool: &'a Pool,
+    ) -> Result<Pin<Box<Self, Pool>>, Error> {
+        // Create a pinned Resolution on the Pool, so that we can make
+        // a stable pointer to the Resolution struct.
+        let mut this = Box::pin_in(
+            Resolution {
+                complete: None,
+                waker: None,
+                pool,
+                ctx: None,
+            },
+            pool.clone(),
+        );
+
+        // Set up the ctx with everything the resolver needs to resolve a
+        // name, and the handler callback which is called on completion.
+        let mut ctx = ResolverCtx::new(resolver.resolver)?;
+        ctx.name = *name;
+        ctx.service = *service;
+        ctx.timeout = resolver.timeout;
+        ctx.set_cancelable(1);
+        ctx.handler = Some(Self::handler);
+
+        {
+            // Safety: Self::handler, Future::poll, and Drop::drop will have
+            // access to &mut Resolution. Nginx is single-threaded and we are
+            // assured only one of those is on the stack at a time, except if
+            // Self::handler wakes a task which polls or drops the Future,
+            // which it only does after use of &mut Resolution is complete.
+            let ptr: &mut Resolution = unsafe { Pin::into_inner_unchecked(this.as_mut()) };
+            ctx.data = ptr as *mut Resolution as *mut c_void;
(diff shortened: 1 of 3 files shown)


pchickey:
Thanks, implemented 1, 2, and 4.

> I'd actually prefer an implementation without futures-channel. There are two heap allocations we can avoid by making this a manual future with proper pinning. And then there's just one extra step to make the code no_std.

Its not obvious to me where these two heap allocations are that we can avoid with proper pinning, and I don't see where it was discussed in the closed PRs. If we're going to tear out the channel, the most obvious improvement to me would be to handle the `receiver.await` future's cancellation by correctly calling `ngx_resolve_name_done`.

bavshin-f5:
> Its not obvious to me where these two heap allocations are that we can avoid with proper pinning, and I don't see where it was discussed in the closed PRs.

`futures_channel::oneshot::{Sender,Receiver}` share an `Arc<Inner>`. That's one allocation and several unnecessary atomic operations.
`Box<ResCtx>` is another allocation that we do for the sole reason of having a callback context with fixed address. A pinned `impl Future` is guaranteed to have a stable address and can be used as a `data` for `ngx_resolver_ctx_t` instead.

And one more heap allocation is in the resulting `Vec` which could be using `Pool` instead.

> If we're going to tear out the channel, the most obvious improvement to me would be to handle the `receiver.await` future's cancellation by correctly calling `ngx_resolve_name_done`. I can't convince myself the current `impl Drop for ResCtx` is ever reachable when ctx is some.

Oh. Right, we consume the `Box` and it is no longer responsible for dropping `rctx`. That is a potential leak.

src/async_/resolver.rs
```diff
@@ -0,0 +1,237 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Error::NoResolver => write!(f, "No resolver configured"),
+            Error::Resolver(err, context) => write!(f, "{err}: resolving `{context}`"),
+            Error::AllocationFailed => write!(f, "Allocation failed"),
+            Error::Unexpected(err) => write!(f, "Unexpected error: {err}"),
+        }
+    }
+}
+impl std::error::Error for Error {}
```
bavshin-f5:
```suggestion
impl core::error::Error for Error {}
```
Our MSRV allows that.

src/async_/resolver.rs
```diff
@@ -0,0 +1,237 @@
...
+// LICENSE file in the root directory of this source tree.
+
+//! Wrapper for the nginx resolver.
+//!
+//! See <https://nginx.org/en/docs/http/ngx_http_core_module.html#resolver>.
+
+use core::ffi::c_void;
+use core::ptr::NonNull;
+use std::boxed::Box;
+use std::fmt;
```
bavshin-f5:
```suggestion
use core::fmt;
```

src/async_/resolver.rs
```diff
@@ -0,0 +1,237 @@
...
+}
+impl std::error::Error for Error {}
+
+/// These cases directly reflect the NGX_RESOLVE_ error codes,
+/// plus a timeout, and a case for an unknown error where a known
+/// NGX_RESOLVE_ should be.
+#[derive(Debug)]
+pub enum ResolverError {
+    /// Format error (NGX_RESOLVE_FORMERR)
+    Format,
```
bavshin-f5:
```suggestion
    FormErr,
```
and NotImp a few lines below.

src/async_/resolver.rs
```diff
@@ -0,0 +1,237 @@
...
+
+//! Wrapper for the nginx resolver.
+//!
+//! See <https://nginx.org/en/docs/http/ngx_http_core_module.html#resolver>.
+
+use core::ffi::c_void;
+use core::ptr::NonNull;
+use std::boxed::Box;
+use std::fmt;
+use std::string::{String, ToString};
```
bavshin-f5:
```suggestion
#[cfg(all(not(feature = "std"), feature = "alloc"))]
use alloc::string::{String, ToString};
#[cfg(feature = "std")]
use std::string::{String, ToString};
```

src/async_/resolver.rs
```diff
@@ -0,0 +1,237 @@
...
+// This source code is licensed under the Apache License, Version 2.0 license found in the
+// LICENSE file in the root directory of this source tree.
+
+//! Wrapper for the nginx resolver.
+//!
+//! See <https://nginx.org/en/docs/http/ngx_http_core_module.html#resolver>.
+
+use core::ffi::c_void;
+use core::ptr::NonNull;
+use std::boxed::Box;
```
bavshin-f5: If the Box is still necessary after all the suggested changes, it could be replaced with `crate::allocator::Box`.

src/async_/resolver.rs
pchickey: Why not unconditonally use alloc::string? `#[cfg(feature = "alloc")] extern crate alloc;` at the crate root is free, and rustc treats any use of std::string as if it were a re-export of alloc::string. We are in a mod that implies feature = "alloc", so we dont have to handle importing neither

bavshin-f5: I was convinced in #111 that this is the right way and adding both `extern crate alloc` and `extern crate std` is not free. Appears that I was wrong.

Please, use alloc::string here as you suggested.

pchickey:
Feedback addressed, I rewrote it as a Future by hand. It still requires one Box because Pin::new requires the inner to have Deref. Tests exist in another repo where @bavshin-f5 can see them. I can port tests to this repo once #198 lands as a basis, but that PR is stuck on an unrelated issue.

I will squash this down once feedback is complete.

src/async_/resolver.rs
```diff
@@ -0,0 +1,316 @@
...
+            // passes a null for the second argument `temp`. A non-null `temp`
+            // provides a fast, non-callback-based path for immediately
+            // returning an addr iff `temp` contains a name which is textual
+            // form of an addr.
+            let ctx = ngx_resolve_start(resolver.as_ptr(), core::ptr::null_mut());
+            NonNull::new(ctx).ok_or(Error::AllocationFailed)?
+        };
+
+        // Create a pinned Resolution on the heap, so that we can make
+        // a stable pointer to the Resolution struct.
```
bavshin-f5: Can we make pinning a caller's responsibility?

If I understand the whole concept of pinning correctly, a value pinned once is no longer movable in a safe code, so taking the address in a method that already requires `Pin<&mut Self>` is sound.

src/async_/resolver.rs
```diff
@@ -0,0 +1,316 @@
...
+                ResolverError::try_from(ret).expect("nonzero, checked above"),
+                name.to_string(),
+            ));
+        }
+
+        Ok(this)
+    }
+
+    // Nginx will call this handler when name resolution completes. If the
+    // result is cached, this could be
```
bavshin-f5: incomplete comment

src/async_/resolver.rs
```diff
@@ -0,0 +1,316 @@
...
+            Err(Error::AllocationFailed)?;
+        }
+        unsafe {
+            addr.sockaddr
+                .cast::<u8>()
+                .copy_to_nonoverlapping(sockaddr.cast(), addr.socklen as usize)
+        };
+
+        let name = unsafe {
+            ngx_str_t::from_bytes(pool.as_ref() as *const _ as *mut _, addr.name.as_bytes())
```
bavshin-f5: This looks unsound. We make *mut ptr from a shared ref, knowing that it will be mutated.

Maybe we need to add `Pool.as_ptr()` or implement `ngx::allocator::TryCloneIn` for `ngx_str_t`.  The latter seems too easy to abuse though.

src/async_/resolver.rs
```diff
@@ -0,0 +1,316 @@
...
+        // Wake last, after all use of &mut Resolution, because wake may
+        // poll Resolution future on current stack.
+        if let Some(waker) = this.waker.take() {
+            waker.wake();
+        }
+    }
+
+    /// Take the results in a ctx and make an owned copy as a
+    /// Result<Vec<ngx_addr_t>, Error>, where both the Vec and internals of
+    /// the ngx_addr_t are allocated on the given Pool
```
bavshin-f5: The comment is not accurate: `Vec` is not allocated on pool.

src/async_/resolver.rs
```diff
@@ -0,0 +1,316 @@
...
+                ResolverError::try_from(s).expect("nonzero, checked above"),
+                ctx.name.to_string(),
+            ));
+        }
+        if ctx.addrs.is_null() {
+            Err(Error::AllocationFailed)?;
+        }
+        let mut out = Vec::new();
+        for i in 0..ctx.naddrs {
+            out.push(Self::copy_resolved_addr(unsafe { ctx.addrs.add(i) }, pool)?);
```
bavshin-f5: Hm...

```diff
diff --git a/src/async_/resolver.rs b/src/async_/resolver.rs
index 2548db3..b26db1f 100644
@@ -240,8 +240,11 @@ impl<'a> Resolution<'a> {
             Err(Error::AllocationFailed)?;
         }
         let mut out = Vec::new();
-        for i in 0..ctx.naddrs {
-            out.push(Self::copy_resolved_addr(unsafe { ctx.addrs.add(i) }, pool)?);
+        if ctx.naddrs > 0 {
+            out = unsafe { core::slice::from_raw_parts(ctx.addrs, ctx.naddrs) }
+                .iter()
+                .map(|x| Self::copy_resolved_addr(x, pool))
+                .collect::<Result<_, _>>()?;
         }
         Ok(out)
     }
@@ -249,14 +252,9 @@ impl<'a> Resolution<'a> {
     /// Take the contents of an ngx_resolver_addr_t and make an owned copy as
     /// an ngx_addr_t, using the Pool for allocation of the internals.
     fn copy_resolved_addr(
-        addr: *mut nginx_sys::ngx_resolver_addr_t,
+        addr: &nginx_sys::ngx_resolver_addr_t,
         pool: &Pool,
     ) -> Result<ngx_addr_t, Error> {
-        let addr = NonNull::new(addr).ok_or(Error::Unexpected(
-            "null ngx_resolver_addr_t in ngx_resolver_ctx_t.addrs".to_string(),
-        ))?;
-        let addr = unsafe { addr.as_ref() };
-
         let sockaddr = pool.alloc(addr.socklen as usize) as *mut nginx_sys::sockaddr;
         if sockaddr.is_null() {
             Err(Error::AllocationFailed)?;
```

This also makes Error::Unexpected unnecessary.

src/async_/resolver.rs
```diff
@@ -0,0 +1,316 @@
...
+
+    /// Take the results in a ctx and make an owned copy as a
+    /// Result<Vec<ngx_addr_t>, Error>, where both the Vec and internals of
+    /// the ngx_addr_t are allocated on the given Pool
+    fn resolve_result(ctx: *mut ngx_resolver_ctx_t, pool: &Pool) -> Res {
+        let ctx = unsafe { ctx.as_ref().unwrap() };
+        let s = ctx.state;
+        if s != 0 {
+            return Err(Error::Resolver(
+                ResolverError::try_from(s).expect("nonzero, checked above"),
```
bavshin-f5: This can be simplified with `impl From<NonZero<isize>> for ResolverError`.
```
if let Some(s) = NonZero::<isize>::new(ctx.state) {
     return Err(Error::Resolver( ResolverError::from(s), ctx.name.to_string()));
}

File: Cargo.toml [hunk 12]
```diff
@@ -42,6 +42,7 @@ async-task = { version = "4.7.1", optional = true }
 lock_api = "0.4.13"
 nginx-sys = { path = "nginx-sys", version = "0.5.0-beta"}
 pin-project-lite = { version = "0.2.16", optional = true }
+futures-channel = { version = "0.3.31", optional = true }
```
bavshin-f5: no longer necessary

src/async_/resolver.rs
pchickey: No, we must pin it here, because we need create a raw pointer out of this Pin<&mut Resolution> in this function at line 187-188.

src/async_/resolver.rs
pchickey: It does look funny, and indicates we need better interfaces on Pool such as `fn as_ptr(&self) -> *mut ngx_pool_t`, as well as improvements to how `ngx_str_t` constructors work. It is not unsound (any more than the whole of nginx is unsound), because all use of these pointers happens across the ffi boundary.

Its not really helpful to think of all of these NonNull<ngx...> wrapper structs in terms of rust shared or unique references, and in particular I believe that all of the impls of AsRef / AsMut for these structs should be eliminated because the resulting types do not fit into the Rust shared/unique reference system. All of these NonNull<ngx...> structs are, under the hood, shared references which permit mutation, which will always be an impedance mismatch because nginx has built in the single-threaded assumption very deeply, and Rust has built in the multi-threaded assumption very deeply. Furthermore, ngx object lifetimes are not representable in the Rust type system - we lose the necessary information at the ffi boundary.

I don't yet have a solution to the problem where reasonable use of these wrapper structs with the ffi requires code that looks like bad Rust - thats essentially the Big Problem with this whole crate and something we need to invest a bunch of time and resources on still, and whatever we come up with is certain to require substantial work throughout this crate and all of its consumer crates. So, I am up for changing this code locally to make it better, but I don't want to start making changes in Pool, ngx_str_t etc as part of this PR.

src/async_/resolver.rs
bavshin-f5: Address-sensitive operations, including call to `ngx_resolve_name()`, can be deferred to `Resolution::poll()`.

src/async_/resolver.rs
pchickey: I misunderstood how we were using the pool here. We will have to make the `Pool` type Copy (which is fine, NonNull<ngx_pool_t> can be shared as much as we want in a single-threaded setting) in order to really make any of these APIs work out as Vec<ngx_addr_t, Pool>. Right now the best we can do is Vec<ngx_addr_t, &'a Pool>, where the 'a doesn't align with anything semantically useful.

src/async_/resolver.rs
bavshin-f5: I believe there is an exact description of this scenario somewhere in the Rust reference: mutating through a pointer derived from a shared reference is undefined behavior. IIUC, it's not related to threading or anything like that, it is a local property of a LLVM representation emitted for this pattern and optimizations permitted by it.

Raw pointer are said to bypass these rules, as long as there is no intermediate step that creates shared ref to the pointee. And pointers as fields of the pointee can be treated as UnsafeCell.
I don't recall if the interpretation is from nomicon or even from discussions in internals, but I've been trying to avoid violating these rules.

NonNull<ngx_...> under this interpretation is sound, because it serves as a more unsafe alternative to UnsafeCell. At no point shared reference is created implicitly, and as_ptr() allows to obtain a *mut pointer in a sound way. NonNull<ngx_...>::as_ref() as *const _ as *mut _ however is not.

src/async_/resolver.rs
pchickey: I don't think that making that change would improve this code. This is a private struct that exists only to provide an implementation for `Resolver::resolve`. If I was to defer ctx initialization and ngx_resolve_name to inside Resolution::poll, I'd have to both hold additional state in `struct Resolution` to carry the ctx, name, and timeout, and add additional complexity to Resolution::poll to do the initialization only on the first call to poll. The result would be less readable because it would move straight-line control flow through these function bodies into a heavy initial case in the poll function,  which is currently quite simple and cleanly separates out the concern of polling for completion, rather than kicking off something that may defer to the event loop. So, those don't really weigh out compared to the goal of changing Box::pin to occur in the caller's scope.

src/async_/resolver.rs
pchickey: OK, I added `pub fn as_ptr(&self) -> *mut ngx_pool_t` to `Pool` and fixed this

src/async_/resolver.rs
```diff
@@ -0,0 +1,337 @@
...
+            // will be called later by nginx when it gets a dns response or a
+            // timeout.
+            let ctx = this
+                .ctx
+                .as_mut()
+                .expect("ctx is Some between Resolution construction and handler");
+            ctx.data = addr.cast();
+            // Does calling the handler inside this call cause UB? We haven't
+            // told rustc that know we have passed &mut Self to the handler by
+            // way of the ctx.
```
bavshin-f5: Indeed, we have two &mut to the same Resolution in the same call stack, and that's not good.
I did a partial revert in https://github.com/nginx/ngx-rust/commit/363b64803bf0bf73a129de3e84490b3c80aeaa55, `git diff 48b4df7..363b648` does not seem to introduce any new unsoundness.

I'm not entirely confident that we are using `Box` correctly here (https://doc.rust-lang.org/stable/std/boxed/index.html#considerations-for-unsafe-code), but it is still better than 100% incorrect code.

PR #204, opened 2025-08-25
test: Link unit tests with nginx library.

ensh63:
### Proposed changes

Some unit tests may use C functions from nginx. To support this scenario, special static library is built. Necessary linker flags are also added. Since these changes are needed for specific unit tests only, new feature "unittest" is defined in `nginx-sys` and `nginx-src` crates.

### Checklist


nginx-sys/build/main.rs
```diff
@@ -286,11 +321,35 @@ pub fn parse_makefile(
...
+            if let Some(lib_dir) = word.strip_prefix("-L") {
+                let value = if lib_dir.is_empty() {
+                    words.next().expect("-L argument")
+                } else {
+                    lib_dir.to_string()
+                };
+                libs.push(LibOption::LibPath(value));
+            } else if let Some(lib) = word.strip_prefix("-l") {
+                let value = if lib.is_empty() {
+                    words.next().expect("-l argument")
```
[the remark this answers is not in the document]
ensh63: This message follows existing format and seems to be rather clear since it describes common flag.

nginx-sys/build/main.rs
```diff
@@ -286,11 +321,35 @@ pub fn parse_makefile(
         }
     }
 
+    fn parse_lib_line(libs: &mut Vec<LibOption>, line: &str) {
+        let mut words = shlex::Shlex::new(line);
+
+        while let Some(word) = words.next() {
+            if let Some(lib_dir) = word.strip_prefix("-L") {
+                let value = if lib_dir.is_empty() {
+                    words.next().expect("-L argument")
```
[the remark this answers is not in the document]
ensh63: Common message format is used; `-L` flag is well-known and does not require additional explanations.

nginx-src/libnginx/libnginx.c
```diff
@@ -0,0 +1,187 @@
...
+
+
+static ngx_int_t libngx_write_temp_conf_file(ngx_cycle_t *cycle,
+    ngx_str_t *data, ngx_str_t *name);
+
+
+ngx_cycle_t *
+libngx_init(u_char *prefix)
+{
+    static ngx_cycle_t  *cycle = NGX_CONF_UNSET_PTR, init_cycle;
```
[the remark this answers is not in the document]
ensh63: NGINX code is not thread-safe generally. Modules and unit tests should follow this pattern communicating with C code.

nginx-src/libnginx/libnginx.c
```diff
@@ -0,0 +1,187 @@
...
+
+    rc = ngx_conf_full_name(cycle, &path->name, 0);
+    if (rc != NGX_OK) {
+        return rc;
+    }
+
+    if (ngx_create_dir(path->name.data, ngx_dir_access(tf.access))
+        == NGX_FILE_ERROR)
+    {
+        return ngx_errno;
```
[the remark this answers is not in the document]
ensh63: Since the only call of this function remaps any error code into `NGX_ERROR`, it is unnecessary to do this inside the function.

src/core/pool.rs
```diff
@@ -326,3 +326,105 @@ impl Pool {
...
+
+    #[link(name = "nginx", kind = "static")]
+    extern "C" {
+        pub fn libngx_init(prefix: *mut nginx_sys::u_char) -> *mut nginx_sys::ngx_cycle_t;
+    }
+
+    #[test]
+    fn test_pool_resize() {
+        unsafe {
+            libngx_init("prefix".as_ptr() as *mut nginx_sys::u_char);
```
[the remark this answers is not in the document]
ensh63: The prefix path her has no any specific meaning; it is just a non-empty string constant which is not used anywhere in this test.

nginx-src/Cargo.toml
```diff
@@ -12,6 +12,10 @@ homepage.workspace = true
 repository.workspace = true
 rust-version.workspace = true
 
+[features]
+# Builds nginx library from vendored sources for unit tests.
+unittest = []
```
bavshin-f5: We can do this unconditionally. `nginx-src` is meant to be used for tests, and building test-only artifacts here is acceptable.

Testing with all supported platforms and nginx versions may be required.

nginx-src/src/lib.rs
```diff
@@ -113,6 +113,12 @@ fn nginx_configure_flags(vendored: &[String]) -> Vec<String> {
         nginx_opts.push(format!("--with-ld-opt={ldflags}"));
     }
 
+    #[cfg(feature = "unittest")]
+    nginx_opts.push(format!(
+        "--add-dynamic-module={}/libnginx",
```
bavshin-f5: `--add-module` is sufficient here

src/core/pool.rs
```diff
@@ -326,3 +326,105 @@ impl Pool {
...
 }
+
+#[cfg(all(test, feature = "vendored"))]
+mod tests {
+
+    use nginx_sys::{ngx_create_pool, ngx_destroy_pool};
+
+    use super::*;
+
+    #[link(name = "nginx", kind = "static")]
```
bavshin-f5: This could be moved to `nginx_sys::unittest::libngx_init`,  a module with test-related utilities conditional under `#[cfg(feature = "unittest")]`.

Another thing to consider is that Rust unit-tests are executed in parallel, and test code will rely on a global state (`ngx_cycle`). Seems like this module should offer a mutex to guard the calls and the test use of nginx configuration. Adding prefix creation via `tempfile` to the wrapper might be a good idea as well. Although `CARGO_TARGET_TMPDIR` should be good enough for now.

If the tests are allowed to enable a feature flag on the current crate, I'd prefer to have this high-level wrapper in the `ngx` itself (because of `std` dependency).

Cargo.toml
```diff
@@ -69,4 +72,5 @@ vendored = ["nginx-sys/vendored"]
 maintenance = { status = "experimental" }
 
 [dev-dependencies]
+nginx-sys = { workspace = true, features = ["unittest"] }
```
bavshin-f5: Should be conditional under `cfg(unix)`, until we can say that `nginx-src` works on Windows.

nginx-src/Cargo.toml
ensh63: removed.

src/core/pool.rs
ensh63: Done with separate package to simplify separation of unittest-related staff from general code.

Cargo.toml
ensh63: Done.

nginx-src/libnginx/config
```diff
@@ -0,0 +1,7 @@
+
+# Copyright (C) Nginx, Inc.
+
+
+ngx_addon_name="libnginx"
+
+CFLAGS="$CFLAGS $ngx_pic_opt"
```
Abhi7739309, line 7: Ok

PR #221, opened 2025-10-22
Trait for HTTP request handler

ensh63:
Trait `HttpRequestHandler<ReturnType>` is defined. This trait allows to implement HTTP request handler with arbitrary return type. Special auxiliary trait `HttpHandlerWrapper<ReturnType>` adds function to convert this return type to `ngx_int_t`. Two conversions are predefined: for `Option<ngx_int_t>` and for `Result<ngx_int_t, E>`. Conversion function may include any desired additional actions common for all handlers. For instance, convertor for `Result<ngx_int_t, E>` logs the error, if any.

Trait `HttpRequestHandler<ReturnType>` also includes associated type for the HTTP module and associated constant for the handler phase.

Handler registration is simplified to a single call of `ngx::http::add_phase_handler::<H: crate::http::HttpHandlerWrapper<R>, R>()` function which is implemented by default.

All examples are modified to use this trait. 

There are some unclear moments:

- Should old handler macro be removed?
- If handler is defined as a separate `struct`, what can be done with data members of this structure? Are they usable at all?
- Is it needed to define a new set of constants for common return values or current approach with `Status` and `into()` is acceptable?

### DIFF
--- a/examples/async.rs
+++ b/examples/async.rs
@@ -6,14 +6,12 @@ use std::time::Instant;
 
 use ngx::core;
 use ngx::ffi::{
-    ngx_array_push, ngx_command_t, ngx_conf_t, ngx_connection_t, ngx_event_t, ngx_http_handler_pt,
-    ngx_http_module_t, ngx_http_phases_NGX_HTTP_ACCESS_PHASE, ngx_int_t, ngx_module_t,
-    ngx_post_event, ngx_posted_events, ngx_posted_next_events, ngx_str_t, ngx_uint_t,
+    ngx_command_t, ngx_conf_t, ngx_connection_t, ngx_event_t, ngx_http_module_t, ngx_int_t,
+    ngx_module_t, ngx_post_event, ngx_posted_events, ngx_posted_next_events, ngx_str_t, ngx_uint_t,
     NGX_CONF_TAKE1, NGX_HTTP_LOC_CONF, NGX_HTTP_LOC_CONF_OFFSET, NGX_HTTP_MODULE, NGX_LOG_EMERG,
 };
-use ngx::http::{self, HttpModule, MergeConfigError};
-use ngx::http::{HttpModuleLocationConf, HttpModuleMainConf, NgxHttpCoreModule};
-use ngx::{http_request_handler, ngx_conf_log_error, ngx_log_debug_http, ngx_string};
+use ngx::http::{self, HttpModule, HttpModuleLocationConf, HttpRequestHandler, MergeConfigError};
+use ngx::{ngx_conf_log_error, ngx_log_debug_http, ngx_string};
 use tokio::runtime::Runtime;
 
 struct Module;
@@ -25,18 +23,10 @@ impl http::HttpModule for Module {
 
     unsafe extern "C" fn postconfiguration(cf: *mut ngx_conf_t) -> ngx_int_t {
         // SAFETY: this function is called with non-NULL cf always
-        let cf = &mut *cf;
-        let cmcf = NgxHttpCoreModule::main_conf_mut(cf).expect("http core main conf");
-
-        let h = ngx_array_push(
-            &mut cmcf.phases[ngx_http_phases_NGX_HTTP_ACCESS_PHASE as usize].handlers,
-        ) as *mut ngx_http_handler_pt;
-        if h.is_null() {
-            return core::Status::NGX_ERROR.into();
-        }
-        // set an Access phase handler
-        *h = Some(async_access_handler);
-        core::Status::NGX_OK.into()
+        let cf = unsafe { &mut *cf };
+        http::add_phase_handler::<AsyncAccessHandler>(cf)
+            .map_or(core::Status::NGX_ERROR, |_| core::Status::NGX_OK)
+            .into()
     }
 }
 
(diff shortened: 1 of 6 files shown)


src/http/request.rs
```diff
@@ -85,6 +86,78 @@ macro_rules! http_variable_get {
...
+    where
+        Self: HttpHandlerWrapper<ReturnType>,
+    {
+        let cmcf = NgxHttpCoreModule::main_conf_mut(cf).expect("http core main conf");
+        let h: *mut ngx_http_handler_pt =
+            unsafe { ngx_array_push(&mut cmcf.phases[Self::PHASE as usize].handlers) as _ };
+        if h.is_null() {
+            return false;
+        }
+        // set an H::PHASE phase handler
```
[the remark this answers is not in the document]
ensh63: Corrected.

nginx-sys/src/http.rs
```diff
@@ -16,3 +16,30 @@ pub const NGX_HTTP_SRV_CONF_OFFSET: usize = offset_of!(ngx_http_conf_ctx_t, srv_
 ///
 /// This is used to access the location configuration context for an HTTP module.
 pub const NGX_HTTP_LOC_CONF_OFFSET: usize = offset_of!(ngx_http_conf_ctx_t, loc_conf);
+
+/// HTTP phases in which a module can register handlers.
+#[repr(u32)]
+pub enum NgxHttpPhases {
```
bavshin-f5: This does not belong in nginx-sys.
Also, please use singular enum name, e.g. `ngx::http::Phase`.

src/http/request.rs
```diff
@@ -85,6 +86,78 @@ macro_rules! http_variable_get {
...
+    const PHASE: NgxHttpPhases;
+    /// The module to which the handler belongs
+    type Module: HttpModule;
+
+    /// The handler function
+    fn handler(request: &mut Request) -> ReturnType;
+    /// Register a request handler for a specified phase.
+    /// This function must be called from the module's `postconfiguration()` function.
+    fn register(cf: &mut ngx_conf_t) -> bool
+    where
```
bavshin-f5: This should not be in this trait, or at least should be protected from overrides.

Ideally it could be a method on a wrapper for `ngx_http_core_main_conf_t`, but for now I'd try a free-standing function.
```rust
mod ngx::http {
pub fn add_phase_handler(cf: &mut ngx_conf_t, handler: impl HttpHandlerWrapper) -> Result<(), AllocError> { ... }`
}
```

src/http/request.rs
```diff
@@ -85,6 +86,78 @@ macro_rules! http_variable_get {
...
+where
+    T: HttpRequestHandler<Option<ngx_int_t>>,
+{
+    fn convert(_r: &Request, res: Option<ngx_int_t>) -> ngx_int_t {
+        res.unwrap_or(NGX_ERROR as _)
+    }
+}
+
+/// Implementation of `HttpHandlerWrapper` for `Result<ngx_int_t, E>`.
+impl<T, E> HttpHandlerWrapper<Result<ngx_int_t, E>> for T
```
bavshin-f5: `Result<HTTPStatus, E>`?

Also, the bounds for `E` can be relaxed to `Display`. I doubt we'll need anything more.

src/http/request.rs
```diff
@@ -85,6 +86,78 @@ macro_rules! http_variable_get {
...
 
+/// Trait for static request handler.
+/// Predefined return types are `Option<ngx_int_t>`
+/// and `Result<ngx_int_t, E>` where `E: ::core::error::Error`.
+/// Handler with `Result<ngx_int_t, E>` return type logs the error message automatically.
+pub trait HttpRequestHandler<ReturnType> {
+    /// The phase in which the handler is invoked
+    const PHASE: NgxHttpPhases;
+    /// The module to which the handler belongs
+    type Module: HttpModule;
```
bavshin-f5: Unused.

examples/async.rs
```diff
@@ -25,17 +23,8 @@ impl http::HttpModule for Module {
...
-        let h = ngx_array_push(
-            &mut cmcf.phases[ngx_http_phases_NGX_HTTP_ACCESS_PHASE as usize].handlers,
-        ) as *mut ngx_http_handler_pt;
-        if h.is_null() {
-            return core::Status::NGX_ERROR.into();
-        }
-        // set an Access phase handler
-        *h = Some(async_access_handler);
+        let cf = unsafe { &mut *cf };
+        AsyncAccessHandler::register(cf);
```
bavshin-f5: Error handling?

nginx-sys/src/http.rs
ensh63: Corrected

src/http/request.rs
ensh63: Bounds relaxed. I'm not sure about `HTTPStatus` here. Probably we need to have more generic status type, but this maybe out of the scope of current PR.

src/http/request.rs
ensh63: Removed.

examples/async.rs
ensh63: Corrected.

src/http/request.rs
ensh63: Corrected.

src/http/request.rs
```diff
@@ -85,6 +87,83 @@ macro_rules! http_variable_get {
...
+/// Trait for static request handler.
+/// Return type must implement [`HttpHandlerReturn`].
+/// There are predefined implementations for `ngx_int_t`, [`Status`], `Option<ngx_int_t>`,
+/// and `Result` with value type implementing [`HttpHandlerReturn`] and error type
+/// implementing `Display`.
+pub trait HttpRequestHandler {
+    /// The phase in which the handler is invoked.
+    const PHASE: HttpPhase;
+    /// The return type of the handler.
+    type ReturnType: HttpHandlerReturn;
```
bavshin-f5: It's a common convention to use `type Output: ...` for cases like this.

src/http/request.rs
```diff
@@ -85,6 +87,83 @@ macro_rules! http_variable_get {
...
+}
+
+impl HttpHandlerReturn for ngx_int_t {
+    #[inline]
+    fn into_ngx_int_t(self, _r: &Request) -> ngx_int_t {
+        self
+    }
+}
+
+impl HttpHandlerReturn for Status {
```
bavshin-f5: Also need an implementation for `HTTPStatus`.

src/http/request.rs
```diff
@@ -85,6 +87,83 @@ macro_rules! http_variable_get {
...
+    #[inline]
+    fn into_ngx_int_t(self, _r: &Request) -> ngx_int_t {
+        self.0
+    }
+}
+
+impl<R, E> HttpHandlerReturn for core::result::Result<R, E>
```
bavshin-f5: `impl<T, E>`? I won't insist, but this would match the `Result` generic args.

src/http/request.rs
```diff
@@ -85,6 +87,83 @@ macro_rules! http_variable_get {
     };
 }
 
+/// Trait for converting handler return types into `ngx_int_t`.
+/// Any desired error handling / logging logic can be implemented in the `into_ngx_int_t` method.
+pub trait HttpHandlerReturn: Sized {
```
bavshin-f5: `IntoStatus` could be a better name for a conversion trait.

src/http/request.rs
```diff
@@ -85,6 +87,83 @@ macro_rules! http_variable_get {
...
+pub trait HttpRequestHandler {
+    /// The phase in which the handler is invoked.
+    const PHASE: HttpPhase;
+    /// The return type of the handler.
+    type ReturnType: HttpHandlerReturn;
+    /// The handler function.
+    fn handler(request: &mut Request) -> Self::ReturnType;
+    /// Handler name for logging purposes.
+    /// `core::any::type_name` is used by default.
+    fn name() -> &'static str {
```
bavshin-f5, line 145: I don't recall if we discussed using `Debug` instead. It should give shorter and more predictable output.

src/http/request.rs
```diff
@@ -85,6 +87,83 @@ macro_rules! http_variable_get {
...
+where
+    R: HttpHandlerReturn,
+    E: Display,
+{
+    #[inline]
+    fn into_ngx_int_t(self, r: &Request) -> ngx_int_t {
+        match self {
+            Ok(val) => val.into_ngx_int_t(r),
+            Err(e) => {
+                crate::ngx_log_error!(NGX_LOG_ERR, r.log(), "{e}",);
```
bavshin-f5: One of the drawbacks of a predefined conversion for Result is that we cannot print a handler name here.

src/http/request.rs
bavshin-f5: Ah, I already wrote about that before.

src/http/request.rs
ensh63: Renamed

src/http/request.rs
ensh63: Renamed

src/http/request.rs
ensh63: `Debug` cannot be used because it is applicable to value, not type itself. Handler value is not required and may not exist.

src/http/request.rs
ensh63: Possible correction added.

src/http/request.rs
ensh63: Renamed

PR #254, opened 2026-02-13
Simple subrequest support

ensh63:
- Typed storage in Pool with cleanup support;
- Request context support for HTTP phase modules;
- Simple version of HTTP subrequest support:
    - subrequest builder;
    - subrequest handler with support of user-supplied function or closure;
    - example of use;
- refactoring for Status conversions

### DIFF
--- a/examples/subrequest.rs
+++ b/examples/subrequest.rs
@@ -1,0 +1,281 @@
+use core::fmt::Display;
+
+use ngx::core::Status;
+use ngx::http::subrequest::{SubRequestBuilder, SubRequestError};
+use ngx::http::{
+    HTTPStatus, HttpModule, HttpModuleLocationConf, HttpPhase, HttpRequestHandler,
+    IntoHandlerStatus, Merge, MergeConfigError, Request, add_phase_handler,
+};
+use ngx::{ngx_log_debug_http, ngx_log_error};
+
+use nginx_sys::{
+    NGX_CONF_TAKE1, NGX_ERROR, NGX_HTTP_LOC_CONF, NGX_HTTP_LOC_CONF_OFFSET, NGX_LOG_ERR,
+    ngx_command_t, ngx_conf_t, ngx_flag_t, ngx_http_complex_value_t, ngx_http_module_t,
+    ngx_http_request_t, ngx_http_send_response, ngx_int_t, ngx_module_t, ngx_str_t, ngx_uint_t,
+};
+
+const NGX_CONF_UNSET_FLAG: ngx_flag_t = nginx_sys::NGX_CONF_UNSET as _;
+
+struct SampleHandler;
+
+enum SampleHandlerError {
+    ContextAllocation,
+    SubRequestCreation(String),
+    SubRequest(ngx_int_t),
+    Response(ngx_int_t),
+}
+
+impl<E> From<SubRequestError<E>> for SampleHandlerError
+where
+    E: Display,
+{
+    fn from(e: SubRequestError<E>) -> Self {
+        SampleHandlerError::SubRequestCreation(e.to_string())
+    }
+}
+
+impl Display for SampleHandlerError {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            SampleHandlerError::ContextAllocation => {
+                write!(f, "context allocation failed")
+            }
+            SampleHandlerError::SubRequestCreation(e) => {
+                write!(f, "subrequest creation failed: {}", e)
+            }
+            SampleHandlerError::SubRequest(rc) => {
+                write!(f, "subrequest failed with return code: {}", rc)
+            }
+            SampleHandlerError::Response(rc) => {
+                write!(f, "response creation failed with return code: {}", rc)
+            }
+        }
+    }
+}
+
+impl IntoHandlerStatus for SampleHandlerError {
+    fn into_handler_status(self, r: &Request) -> ngx_int_t {
+        ngx_log_error!(NGX_LOG_ERR, r.log(), "subrequest example: {self}");
+        Status::NGX_ERROR.into()
+    }
+}
+
+impl HttpRequestHandler for SampleHandler {
+    const PHASE: HttpPhase = HttpPhase::Access;
+    type Output = Result<Status, SampleHandlerError>;
+
+    fn handler(request: &mut Request) -> Self::Output {
+        let co = Module::location_conf(request).expect("module config is none");
+        ngx_log_debug_http!(request, "subrequest module enabled: {}", co.enable);
+
+        if co.enable != 1 {
+            return Ok(Status::NGX_DECLINED);
+        }
+
+        let rptr: *mut ngx_http_request_t = request.as_mut();
+
+        match SRCtx::get(request) {
+            Some(ctx) => ctx.rc.map_or(
+                // `ctx` has been created but not filled yet - subrequest is still in progress
+                Ok(Status::NGX_AGAIN),
+                // `ctx` has been created and filled - subrequest is completed
+                |rc| {
+                    let status = ctx.status.0;
+                    let msg = format!("subrequest completed with HTTP status: {status}, rc: {rc}");
+                    ngx_log_debug_http!(request, "{msg}");
+
+                    if status >= nginx_sys::NGX_HTTP_SPECIAL_RESPONSE as _ {
+                        Ok(Status::from(ctx.status))
+                    } else if rc == nginx_sys::NGX_OK as _ && ctx.out.is_some() {
+                        let outbuf = unsafe { &*ctx.out.unwrap().buf };
+                        let mut ct = ctx.ct;
+                        let mut cv: ngx_http_complex_value_t = unsafe { core::mem::zeroed() };
+                        cv.value = ngx_str_t {
+                            len: unsafe { outbuf.last.offset_from(outbuf.pos) } as _,
+                            data: outbuf.pos as _,
+                        };
+                        let resp_rc = unsafe {
+                            ngx_http_send_response(rptr, status, &raw mut ct, &raw mut cv)
+                        };
+                        if resp_rc == nginx_sys::NGX_OK as _ {
+                            Ok(Status::from(ctx.status))
+                        } else {
+                            Err(SampleHandlerError::Response(resp_rc))
+                        }
+                    } else if rc == nginx_sys::NGX_OK as _ {
+                        Ok(Status::from(ctx.status))
+                    } else if let Ok(http_status) = HTTPStatus::try_from(rc) {
+                        Ok(Status::from(http_status))
+                    } else {
+                        Err(SampleHandlerError::SubRequest(rc))
+                    }
+                },
+            ),
+            None => {
+                if SRCtx::create(request).is_some() {
+                    let uri: &str = co.uri.to_str().unwrap_or("/proxy");
+
+                    SubRequestBuilder::new(request, uri)?
+                        .args("arg1=val1&arg2=val2")?
+                        .init(|sr| {
+                            ngx_log_debug_http!(
+                                sr,
+                                "initializing subrequest with URI: {}",
+                                sr.path()
+                            );
+                            sr.add_header_in("X-SubRequest", "1").ok_or("cannot add header")
+                        })
+                        .handler(sr_handler)
+                        .in_memory()
+                        .waited()
+                        .build()?;
+
+                    Ok(Status::NGX_AGAIN)
+                } else {
+                    Err(SampleHandlerError::ContextAllocation)
+                }
+            }
+        }
+    }
+}
+
+struct SRCtx<'r> {
+    rc: Option<ngx_int_t>,
+    status: HTTPStatus,
+    out: Option<&'r nginx_sys::ngx_chain_t>,
+    ct: ngx_str_t,
+}
+
+impl SRCtx<'_> {
+    fn create(request: &mut Request) -> Option<&mut Self> {
+        let ctx_ref = unsafe { request.pool().allocate_with_cleanup(Self::default).ok()?.as_mut() };
+        request.set_module_ctx(ctx_ref as *mut _ as _, Module::module());
+        Some(ctx_ref)
+    }
+
+    fn get(request: &Request) -> Option<&Self> {
+        request.get_module_ctx::<Self>(Module::module())
+    }
+
+    fn get_mut(request: &mut Request) -> Option<&mut Self> {
+        request.get_module_ctx_mut::<Self>(Module::module())
+    }
+}
+
+impl Default for SRCtx<'_> {
+    fn default() -> Self {
+        Self { rc: None, status: HTTPStatus(NGX_ERROR as _), out: None, ct: ngx_str_t::empty() }
+    }
+}
+
+fn sr_handler(r: &mut Request, mut rc: ngx_int_t) -> ngx_int_t {
+    let newctx = SRCtx {
+        rc: Some(rc),
+        status: r.status(),
+        // SAFETY: `r.as_ref().out` is valid as long as the main request is not finalized,
+        // and the subrequest is always finalized before the main request.
+        out: core::ptr::NonNull::new(r.as_ref().out).map(|out| unsafe { out.as_ref() }),
+        ct: r.as_ref().headers_out.content_type,
+    };
+    if let Some(ctx) = SRCtx::get_mut(r.main_mut()) {
+        *ctx = newctx;
+    } else {
+        ngx_log_error!(nginx_sys::NGX_LOG_ERR, r.log(), "subrequest: context not found");
+        rc = NGX_ERROR as _;
+    }
+    rc
+}
+
+static NGX_HTTP_SUBREQUEST_MODULE_CTX: ngx_http_module_t = ngx_http_module_t {
+    preconfiguration: None,
+    postconfiguration: Some(Module::postconfiguration),
+    create_main_conf: None,
+    init_main_conf: None,
+    create_srv_conf: None,
+    merge_srv_conf: None,
+    create_loc_conf: Some(Module::create_loc_conf),
+    merge_loc_conf: Some(Module::merge_loc_conf),
+};
+
+#[cfg(feature = "export-modules")]
+ngx::ngx_modules!(ngx_http_subrequest_module);
+
+#[used]
+#[allow(non_upper_case_globals)]
+#[cfg_attr(not(feature = "export-modules"), unsafe(no_mangle))]
+pub static mut ngx_http_subrequest_module: ngx_module_t = ngx_module_t {
+    ctx: &raw const NGX_HTTP_SUBREQUEST_MODULE_CTX as _,
+    commands: unsafe { &raw mut NGX_HTTP_SUBREQUEST_COMMANDS[0] },
+    type_: nginx_sys::NGX_HTTP_MODULE as _,
+    ..ngx_module_t::default()
+};
+
+struct Module;
+
+impl HttpModule for Module {
+    fn module() -> &'static ngx_module_t {
+        unsafe { &*::core::ptr::addr_of!(ngx_http_subrequest_module) }
+    }
+
+    unsafe extern "C" fn postconfiguration(cf: *mut ngx_conf_t) -> ngx_int_t {
+        // SAFETY: this function is called with non-NULL cf always
+        let cf = unsafe { &mut *cf };
+        add_phase_handler::<SampleHandler>(cf)
+            .map_or(nginx_sys::NGX_ERROR as _, |_| nginx_sys::NGX_OK as _)
+    }
+}
+
+#[derive(Debug)]
+struct ModuleConfig {
+    enable: ngx_flag_t,
+    uri: ngx_str_t,
+}
+
+impl Default for ModuleConfig {
+    fn default() -> Self {
+        Self { enable: NGX_CONF_UNSET_FLAG, uri: ngx_str_t::empty() }
+    }
+}
+
+impl Merge for ModuleConfig {
+    fn merge(&mut self, prev: &ModuleConfig) -> Result<(), MergeConfigError> {
+        if self.enable == NGX_CONF_UNSET_FLAG {
+            if prev.enable != NGX_CONF_UNSET_FLAG {
+                self.enable = prev.enable;
+            } else {
+                self.enable = 0;
+            }
+        }
+        if self.uri.data.is_null() {
+            self.uri = prev.uri;
+        }
+        if self.enable == 1 && self.uri.data.is_null() {
+            self.uri = ngx::ngx_string!("/proxy");
+        }
+        Ok(())
+    }
+}
+
+unsafe impl HttpModuleLocationConf for Module {
+    type LocationConf = ModuleConfig;
+}
+
+static mut NGX_HTTP_SUBREQUEST_COMMANDS: [ngx_command_t; 3] = [
+    ngx_command_t {
+        name: ngx::ngx_string!("subrequest"),
+        type_: (NGX_HTTP_LOC_CONF | NGX_CONF_TAKE1) as ngx_uint_t,
+        set: Some(nginx_sys::ngx_conf_set_flag_slot),
+        conf: NGX_HTTP_LOC_CONF_OFFSET,
+        offset: core::mem::offset_of!(ModuleConfig, enable),
+        post: core::ptr::null_mut(),
+    },
+    ngx_command_t {
+        name: ngx::ngx_string!("subrequest_uri"),
+        type_: (NGX_HTTP_LOC_CONF | NGX_CONF_TAKE1) as ngx_uint_t,
+        set: Some(nginx_sys::ngx_conf_set_str_slot),
+        conf: NGX_HTTP_LOC_CONF_OFFSET,
+        offset: core::mem::offset_of!(ModuleConfig, uri),
+        post: core::ptr::null_mut(),
+    },
+    ngx_command_t::empty(),
+];
--- a/examples/t/subrequest.t
+++ b/examples/t/subrequest.t
@@ -1,0 +1,79 @@
+#!/usr/bin/perl
+
+# (C) Nginx, Inc
+
+# Tests for ngx-rust example modules.
+
+###############################################################################
+
+use warnings;
+use strict;
+
+use Test::More;
+
+BEGIN { use FindBin; chdir($FindBin::Bin); }
+
+use lib 'lib';
+use Test::Nginx;
+
+###############################################################################
+
+select STDERR; $| = 1;
+select STDOUT; $| = 1;
+
+my $t = Test::Nginx->new()->has(qw/http proxy/)->plan(2)
+    ->write_file_expand('nginx.conf', <<'EOF');
+
+%%TEST_GLOBALS%%
+
+daemon off;
+
+events {
+}
+
+http {
+    %%TEST_GLOBALS_HTTP%%
+
+    server {
+        listen       127.0.0.1:8080;
+        server_name  localhost;
+
+        location / {
+            subrequest on;
+        }
+
+        location /non_existing {
+            subrequest on;
+            subrequest_uri /non_existing_upstream;
+        }
+
+        location /proxy {
+            internal;
+            proxy_pass http://127.0.0.1:8081;
+        }
+    }
+
+    server {
+        listen      127.0.0.1:8081;
+        server_name localhost;
+
+        location / {
+            set $reply 'Invalid';
+            if ($http_x_subrequest) {
+                set $reply 'Hello from backend';
+            }
+            return 200 $reply;
+        }
+    }
+}
+
+EOF
+
+$t->write_file('index.html', '');
+$t->run();
+
+like(http_get('/'),
+    qr/200 OK.*Hello from backend/s,
+    'subrequest');
+like(http_get('/non_existing'), qr/404 Not Found/s,
+    'subrequest to non-existing upstream');
--- a/src/core/pool.rs
+++ b/src/core/pool.rs
@@ -5,7 +5,7 @@ use core::ptr::{self, NonNull};
 
 use nginx_sys::{
     NGX_ALIGNMENT, ngx_buf_t, ngx_create_temp_buf, ngx_palloc, ngx_pcalloc, ngx_pfree,
-    ngx_pmemalign, ngx_pnalloc, ngx_pool_cleanup_add, ngx_pool_t,
+    ngx_pmemalign, ngx_pnalloc, ngx_pool_cleanup_add, ngx_pool_cleanup_t, ngx_pool_t,
 };
 
 use crate::allocator::{AllocError, Allocator, dangling_for_layout};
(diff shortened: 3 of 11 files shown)


src/http/subrequest.rs
```diff
@@ -0,0 +1,170 @@
...
+                &raw mut sr_ptr,
+                Box::as_mut_ptr(&mut psr),
+                self.flags as ngx_uint_t,
+            );
+
+            if rc != nginx_sys::NGX_OK as _ {
+                return Err(SubRequestError::CreationFailed);
+            }
+        }
+
```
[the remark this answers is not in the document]
ensh63: Corrected

src/http/subrequest.rs
```diff
@@ -0,0 +1,170 @@
...
+    pub fn build<H, O>(&self, request: &mut Request, handler: H) -> Result<(), SubRequestError>
+    where
+        H: FnMut(&mut Request, ngx_int_t) -> O,
+        O: IntoHandlerStatus,
+    {
+        let ctx = Box::try_new_in(handler, request.pool())
+            .map_err(|_| SubRequestError::RequestAllocFailed)?;
+        let mut ctx = Box::into_pin(ctx);
+
+        self.build_internal(request, ctx.as_mut())
```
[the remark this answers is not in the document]
ensh63: corrected

src/http/request.rs
```diff
@@ -196,12 +214,40 @@ impl Request {
...
+        unsafe { &*r.cast::<Request>() }
+    }
+
     /// Is this the main request (as opposed to a subrequest)?
     pub fn is_main(&self) -> bool {
         let main = self.0.main.cast();
         core::ptr::eq(self, main)
     }
 
+    /// Get a mutable reference to the main request if this is a subrequest, otherwise `None`.
```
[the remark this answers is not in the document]
ensh63: Corrected

src/http/status.rs
```diff
@@ -48,35 +40,42 @@ impl From<HTTPStatus> for ngx_uint_t {
...
-    /// Convets a u16 to a status code.
     #[inline]
-    pub fn from_u16(src: u16) -> Result<HTTPStatus, InvalidHTTPStatusCode> {
-        if !(100..600).contains(&src) {
-            return Err(InvalidHTTPStatusCode::new());
+    fn try_from(value: usize) -> Result<Self, Self::Error> {
+        if !(100..600).contains(&value) {
+            return Err(InvalidHTTPStatusCode);
         }
+        Ok(HTTPStatus(value))
```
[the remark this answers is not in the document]
ensh63: The range check in lines 48--50 eliminates the possibility of erroneous conversion for any huge value.

src/http/subrequest.rs
```diff
@@ -0,0 +1,158 @@
...
+    r: *mut ngx_http_request_t,
+    data: *mut c_void,
+    rc: ngx_int_t,
+) -> ngx_int_t
+where
+    H: Fn(&mut Request, ngx_int_t) -> O,
+    O: IntoHandlerStatus,
+{
+    let request = unsafe { Request::from_ngx_http_request(r) };
+    let handler = unsafe { &mut *(data as *mut H) };
```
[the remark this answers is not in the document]
ensh63: corrected

examples/subrequest.rs
```diff
@@ -0,0 +1,244 @@
...
+            );
+            ngx_log_debug_http!(request, "{msg}");
+
+            if ctx.status.0 >= nginx_sys::NGX_HTTP_SPECIAL_RESPONSE as _ {
+                return Ok(Status::from(ctx.status));
+            }
+
+            if rc == nginx_sys::NGX_OK as _ {
+                let mut cv: ngx_http_complex_value_t = unsafe { core::mem::zeroed() };
+                cv.value = unsafe { ngx_str_t::from_str(request.pool().as_ptr(), &msg) };
```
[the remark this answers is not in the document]
ensh63: Refactored; now there is no allocation here.

examples/subrequest.rs
```diff
@@ -0,0 +1,244 @@
...
+
+        let request_ptr: *mut ngx_http_request_t = request.as_mut();
+
+        if let Some(ctx) = SRCtx::get(request) {
+            if ctx.rc.is_none() {
+                return Ok(Status::NGX_AGAIN);
+            }
+
+            let rc = ctx.rc.unwrap();
+
```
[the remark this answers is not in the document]
ensh63: Refactored

src/http/request_context.rs
```diff
@@ -0,0 +1,40 @@
...
+        let ctx_ref = unsafe { request.pool().allocate_with_cleanup(f)?.as_mut() };
+        request.set_module_ctx(ctx_ref as *mut _ as _, Module::module());
+        Some(ctx_ref)
+    }
+
+    /// Removes the context associated with the given request.
+    fn remove(request: &mut Request) {
+        if let Some(ctx_ptr) = request.get_module_ctx::<Self>(Module::module()) {
+            request.set_module_ctx(core::ptr::null_mut(), Module::module());
+            unsafe { request.pool().remove(ctx_ptr as *const Self) };
```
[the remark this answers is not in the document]
ensh63: Corrected

src/core/pool.rs
```diff
@@ -206,25 +211,97 @@ impl Pool {
...
+    /// the closure is not called.
+    ///
+    /// Returns `Some(NonNull<T>)` if the allocation and cleanup handler addition are successful,
+    /// or `None` if allocation fails.
     ///
     /// # Safety
     /// This function is marked as unsafe because it involves raw pointer manipulation.
-    unsafe fn add_cleanup_for_value<T>(&self, value: *mut T) -> Result<(), ()> {
-        let cln = unsafe { ngx_pool_cleanup_add(self.0.as_ptr(), 0) };
+    /// The caller must ensure that the returned pointer is used safely.
```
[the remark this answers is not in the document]
ensh63: Corrected

src/core/pool.rs
```diff
@@ -128,6 +128,11 @@ impl AsMut<ngx_pool_t> for Pool {
     }
 }
 
+// Wrapper to create an unique value type
+struct Item<T: Sized> {
```
bavshin-f5:
```rust
/// Private wrapper to prevent API users from storing additional instances of unique.
struct UniqueItem<T...>(T)
```
Also, do we need the Sized bound here and in remove_unique? It seems redundant, because the only thing we have to block is an attempt to allocate unsized.

src/core/pool.rs
```diff
@@ -272,18 +314,136 @@ impl Pool {
     ///
     /// Returns a typed pointer to the allocated memory if successful, or a null pointer if
     /// allocation or cleanup handler addition fails.
-    pub fn allocate<T>(&self, value: T) -> *mut T {
+    pub fn allocate<T: Sized>(&self, value: T) -> *mut T {
```
bavshin-f5, line 288: I believe we can drop this method in favor of `allocate_with_cleanup`. It's poorly named, conflicts with `Allocator::allocate` and offers a bad, unsafe API that doesn't encourage checking the result.

It would force us to convert several call sites to a more verbose construction, but I actually consider this a benefit:
```diff
diff --git a/src/http/module.rs b/src/http/module.rs
index 7a9292c..3310d8d 100644
--- a/src/http/module.rs
+++ b/src/http/module.rs
@@ -79,7 +79,10 @@ pub trait HttpModule {
     {
         unsafe {
             let pool = Pool::from_ngx_pool((*cf).pool);
-            pool.allocate::<Self::MainConf>(Default::default()) as *mut c_void
+            pool.allocate_with_cleanup(Self::MainConf::default)
+                .map(ptr::NonNull::as_ptr)
+                .unwrap_or(ptr::null_mut())
+                .cast()
         }
     }
 
```

Here, we avoid copying temporary (by passing the `default` constructor to `allocate_with_cleanup`), and explicitly handle the `None`. It might even result in a similar assembly output.

src/core/pool.rs
```diff
@@ -272,18 +314,136 @@ impl Pool {
...
+
+    /// Gets the unique value of type `T` from the memory pool, or allocates and initializes it
+    /// using the provided function if it does not exist.
+    ///
+    /// This ensures only one value of type `T` exists in the pool. If a value already exists,
+    /// it is returned and the function `f` is not called. If no value exists, a new one is
+    /// created using `f` and stored with a cleanup handler. If allocation fails, `f` is not called.
+    ///
+    /// Returns a mutable reference to the value if successful, or `None` if allocation fails.
+    pub fn get_or_add_unique<T: Sized, F: FnOnce() -> T>(&mut self, f: F) -> Option<&mut T> {
```
bavshin-f5: Looking at the `std` name conventions, this would be `get_or_insert_unique_with`.

src/core/pool.rs
```diff
@@ -200,25 +205,62 @@ impl Pool {
...
 
-        Ok(())
+    /// Searches for a cleanup handler in the pool's cleanup chain.
+    ///
+    /// Returns `Some(NonNull<ngx_pool_cleanup_t>)` if a cleanup handler is found for type `T`,
+    /// or `None` if no matching handler is found.
+    ///
+    /// # Safety
+    /// This function is marked as unsafe because it involves raw pointer manipulation
+    /// and traverses the nginx cleanup chain structure.
```
bavshin-f5: If the method has no safety invariants or other restrictions on the _caller_, it's not unsafe.

If the method has such preconditions or limitations, it must list these instead of an useless "marked as unsafe because it involves raw pointer manipulation" placeholder.

src/core/pool.rs
```diff
@@ -332,6 +492,8 @@ impl Pool {
 /// * `data` - A raw pointer to the value of type `T` to be cleaned up.
 unsafe extern "C" fn cleanup_type<T>(data: *mut c_void) {
     unsafe {
-        ptr::drop_in_place(data as *mut T);
+        if !data.is_null() {
```
bavshin-f5: Is it possible to call this with NULL?

src/core/pool.rs
```diff
@@ -272,18 +314,136 @@ impl Pool {
...
+        self.remove_cleanup_handler(|cln| cln.handler == Some(cleanup_type::<T>)).map(|cln| {
+            unsafe { cln.handler.unwrap()(cln.data) };
+        })
+    }
+
+    /// Internal method to find and remove a cleanup handler matching a predicate.
+    ///
+    /// Returns `Some(&ngx_pool_cleanup_t)` if a matching handler is found and removed,
+    /// or `None` if not found.
+    fn remove_cleanup_handler(
```
bavshin-f5: `remove_cleanup_if`?

src/core/pool.rs
```diff
@@ -272,18 +314,136 @@ impl Pool {
...
+    }
+
+    /// Runs the cleanup handler for a unique value and removes it.
+    ///
+    /// This value must have been previously allocated with [`Pool::get_or_add_unique`].
+    ///
+    /// Returns `Some(())` if the value was successfully removed,
+    /// or `None` if the value was not found.
+    pub fn remove_unique<T: Sized>(&mut self) -> Option<()> {
+        // SAFETY: comparing function pointers is generally unreliable, but in this specific
```
bavshin-f5: Safety comment without a corresponding unsafe block.

src/core/pool.rs
```diff
@@ -272,18 +314,136 @@ impl Pool {
...
+            cln.handler == Some(cleanup_type::<T>) && core::ptr::eq(cln.data, value as _)
+        })
+        .map(|cln| {
+            unsafe { cln.handler.unwrap()(cln.data) };
+        })
+    }
+
+    /// Runs the cleanup handler for a unique value and removes it.
+    ///
+    /// This value must have been previously allocated with [`Pool::get_or_add_unique`].
```
bavshin-f5: What would happen if it wasn't?

src/core/pool.rs
```diff
@@ -272,18 +314,136 @@ impl Pool {
...
+    /// Runs the cleanup handler for a value and removes it.
+    ///
+    /// Returns `Some(())` if the value was successfully removed,
+    /// or `None` if the value was not found.
+    ///
+    /// # Safety
+    /// The caller must ensure that `value` is a valid pointer to a value that has an
+    /// associated cleanup handler in the pool.
+    pub unsafe fn remove<T: Sized>(&mut self, value: *const T) -> Option<()> {
+        // SAFETY: comparing function pointers is generally unreliable, but in this specific
```
bavshin-f5: Safety comment without a corresponding unsafe block

src/core/pool.rs
```diff
@@ -272,18 +314,136 @@ impl Pool {
...
+        // case we can assume that the same function pointer was used when adding the cleanup
+        // handler.
+        #[allow(unpredictable_function_pointer_comparisons)]
+        self.remove_cleanup_handler(|cln| {
+            cln.handler == Some(cleanup_type::<T>) && core::ptr::eq(cln.data, value as _)
```
bavshin-f5: `ptr::addr_eq(cln.data, value)`

src/core/pool.rs
```diff
@@ -272,18 +314,136 @@ impl Pool {
...
+                None => ptr::null_mut(),
+                Some(mut ptr) => ptr.as_mut(),
+            }
+        }
+    }
+
+    /// Gets the unique value of type `T` from the memory pool, or allocates and initializes it
+    /// using the provided function if it does not exist.
+    ///
+    /// This ensures only one value of type `T` exists in the pool. If a value already exists,
```
bavshin-f5: It doesn't,  it's still possible to do `allocate_with_cleanup(T::default)`.

src/core/pool.rs
```diff
@@ -200,25 +205,62 @@ impl Pool {
...
+    /// Searches for a cleanup handler in the pool's cleanup chain.
+    ///
+    /// Returns `Some(NonNull<ngx_pool_cleanup_t>)` if a cleanup handler is found for type `T`,
+    /// or `None` if no matching handler is found.
+    ///
+    /// # Safety
+    /// This function is marked as unsafe because it involves raw pointer manipulation
+    /// and traverses the nginx cleanup chain structure.
+    unsafe fn cleanup_lookup<T: Sized>(&self) -> Option<NonNull<ngx_pool_cleanup_t>> {
+        let mut cln = (unsafe { *self.0.as_ptr() }).cleanup;
```
bavshin-f5: Slightly less unsafe:
```rust
let mut cln = NonNull::new(self.as_ref().cleanup)?;
loop {
    #[allow(unpredictable_function_pointer_comparisons)]
    if unsafe { cln.as_ref().handler } == Some(cleanup_type::<T>) {
        return Some(cln);
    }
                                                                    
    cln = NonNull::new(unsafe { cln.as_ref() }.next)?;
}
```

However, we don't modify the result of this function and don't really need `NonNull`:
```rust
fn get_cleanup<T>(&self) -> Option<&ngx_pool_cleanup_t> {
    let mut cln = unsafe { self.as_ref().cleanup.as_ref() }?;
    loop {
       #[allow(unpredictable_function_pointer_comparisons)]
        if cln.handler == Some(cleanup_type::<T>) {
            return Some(cln);
        }

        cln = unsafe { cln.next.as_ref() }?;
    }
}
```

src/core/pool.rs
```diff
@@ -200,25 +205,62 @@ impl Pool {
...
-        Ok(())
+    /// Searches for a cleanup handler in the pool's cleanup chain.
+    ///
+    /// Returns `Some(NonNull<ngx_pool_cleanup_t>)` if a cleanup handler is found for type `T`,
+    /// or `None` if no matching handler is found.
+    ///
+    /// # Safety
+    /// This function is marked as unsafe because it involves raw pointer manipulation
+    /// and traverses the nginx cleanup chain structure.
+    unsafe fn cleanup_lookup<T: Sized>(&self) -> Option<NonNull<ngx_pool_cleanup_t>> {
```
bavshin-f5: `get_cleanup`?

src/core/pool.rs
```diff
@@ -200,25 +205,62 @@ impl Pool {
...
+        &self,
+        f: F,
+    ) -> Option<NonNull<T>> {
+        let cln = unsafe { ngx_pool_cleanup_add(self.0.as_ptr(), mem::size_of::<T>()) };
         if cln.is_null() {
-            return Err(());
+            return None;
         }
-
         unsafe {
```
bavshin-f5: The unsafe scope here can be easily reduced:
```rust
let cln = unsafe { ngx_pool_cleanup_add(self.0.as_ptr(), mem::size_of::<T>()).as_mut()? };
if cln.data.is_null() {
    cln.data = ptr::from_mut(cln).cast();
};
cln.handler = Some(cleanup_type::<T>);
unsafe { ptr::write(cln.data.cast(), f()) };
NonNull::new(cln.data.cast())
```

bavshin-f5 (changes requested):

src/core/pool.rs
ensh63: Corrected

src/core/pool.rs
ensh63: Agree, but this is probably out of scope of this PR. Old allocation methods should be overlooked in a separate commit.

src/core/pool.rs
ensh63: No, this is a leftover. Removed.

src/core/pool.rs
ensh63: Corrected.

src/core/pool.rs
ensh63: Corrected

src/core/pool.rs
ensh63: Corrected

src/core/pool.rs
ensh63: Renamed

src/core/pool.rs
ensh63: Corrected

src/core/pool.rs
ensh63: Renamed

src/core/pool.rs
ensh63: Corrected

src/core/pool.rs
ensh63: Corrected

src/core/pool.rs
ensh63: Nothing; the function is no-op if cleanup handler is not found.
Comment corrected.

src/core/pool.rs
ensh63: Renamed

src/core/pool.rs
ensh63: Comment corrected

src/core/pool.rs
```diff
@@ -128,6 +128,11 @@ impl AsMut<ngx_pool_t> for Pool {
     }
 }
 
+/// Private wrapper to prevent API users from storing additional instances of unique.
+struct UniqueItem<T> {
+    value: T,
+}
```
bavshin-f5: This doesn't have to be a real struct; please, convert to newtype (`struct UniqueItem<T>(T)`);

src/core/pool.rs
```diff
@@ -200,25 +205,52 @@ impl Pool {
...
-    unsafe fn add_cleanup_for_value<T>(&self, value: *mut T) -> Result<(), ()> {
-        let cln = unsafe { ngx_pool_cleanup_add(self.0.as_ptr(), 0) };
-        if cln.is_null() {
-            return Err(());
+    /// The returned pointer must not outlive the pool, must not be freed manually
+    /// (as it has a cleanup handler), and must not be accessed after the pool is destroyed.
+    pub unsafe fn allocate_with_cleanup<T: Sized, F: FnOnce() -> T>(
+        &self,
+        f: F,
+    ) -> Option<NonNull<T>> {
```
bavshin-f5: As I suggested offline, returning `Result<NonNull<T>, AllocError>` here can make subsequent error handling patterns simpler.

src/http/subrequest.rs
```diff
@@ -0,0 +1,201 @@
...
+    /// Indicates that the post subrequest allocation failed.
+    PostRequestAllocFailed,
+    /// Indicates that the URI allocation failed.
+    UriAllocFailed,
+    /// Indicates that the arguments allocation failed.
+    ArgsAllocFailed,
+    /// Indicates that the subrequest creation failed.
+    CreationFailed,
+    /// Indicates that the subrequest modification failed.
+    ModificationFailed(String),
```
bavshin-f5: Stringifying the nested error might be a bit excessive; we want to reduce allocation on error paths.

I also don't think we need to track all possible allocation failures separately.

I'd suggest a simpler error type: 
```rust
/// An error type for subrequest operations.
#[derive(Debug)]
pub enum SubRequestError<E = core::convert::Infallible> {
    /// Indicates that the subrequest allocation failed.
    Alloc,
    /// Indicates that the subrequest creation failed.
    Create,
    /// Indicates that the subrequest modification failed.
    Modify(E),
}

impl<E> From<AllocError> for SubRequestError<E> {
    fn from(_: AllocError) -> Self {
        Self::Alloc
    }
}
```

src/http/request.rs
```diff
@@ -104,6 +107,23 @@ where
     }
 }
 
+impl<T, E> IntoHandlerStatus for Result<T, E>
```
bavshin-f5, line 108: I believe we already discussed this: blanket IntoHandlerStatus implementation for Result in the library blocks any possibility of redefinition by users. Is it really necessary?

src/http/request.rs
```diff
@@ -195,12 +215,42 @@ impl Request {
...
     pub fn is_main(&self) -> bool {
         let main = self.0.main.cast();
         core::ptr::eq(self, main)
     }
 
+    /// Get a mutable reference to the main request.
+    ///
+    /// If this is already the main request, returns `self`; otherwise returns
+    /// a mutable reference to the associated main request.
+    pub fn get_main_mut(&mut self) -> &mut Request {
```
bavshin-f5: To be consistent with the existing methods: `main`, `main_mut`, `status`...

src/http/request.rs
ensh63: User-defined redefinitions are impossible in any case, because both `Result` and `IntoHandlerStatus` are not defined by user. But anyway, I prepared variant without this implementation. Not sure, if it looks better...

bavshin-f5: > User-defined redefinitions are impossible in any case, because both Result and IntoHandlerStatus are not defined by user.

That's fair. There are exceptions for some types, but not for the Result.

How about this approach?
```rust
// src/http/request.rs:
impl<T, E> IntoHandlerStatus for Result<T, E>
where
    T: IntoHandlerStatus,
    E: IntoHandlerStatus,
{
    fn into_handler_status(self, r: &Request) -> ngx_int_t {
        match self {
            Ok(val) => val.into_handler_status(r),
            Err(err) => err.into_handler_status(r),
        }
    }
}
// examples/subrequest.rs
impl IntoHandlerStatus for SampleHandlerError {
    fn into_handler_status(self, r: &Request) -> ngx_int_t {
        ngx_log_error!(NGX_LOG_ERR, r.log(), "subrequest example: {self}");
        Status::NGX_ERROR.into()
    }
}
```

ensh63: Looks good. I applied these changes with another, but equivalent, implementation for `into_handler_status()` for `Result` to make it similar to existing implementation for `Option`.

src/http/subrequest.rs
```diff
@@ -0,0 +1,192 @@
...
+        self.build_ext(request, SR_MODIFIER_NO_OP, Some(handler))
+    }
+}
+
+type SimpleSubRequestModifier = Option<fn(&mut Request) -> Result<(), core::convert::Infallible>>;
+type SimpleSubRequestHandler = Option<fn(&mut Request, ngx_int_t) -> ngx_int_t>;
+/// A no-op modifier function for subrequests.
+pub const SR_MODIFIER_NO_OP: SimpleSubRequestModifier = None;
+/// A no-op handler function for subrequests.
+pub const SR_HANDLER_NO_OP: SimpleSubRequestHandler = None;
```
bavshin-f5: It feels like introducing all of this (with private types) instead of simply writing `None` is excessive.

src/http/subrequest.rs
```diff
@@ -0,0 +1,192 @@
...
+            .ok_or(SubRequestError::Alloc)?;
+        self.args = Some(args);
+        Ok(self)
+    }
+
+    /// Sets the subrequest to be in-memory.
+    pub fn in_memory(mut self) -> Self {
+        self.flags |= nginx_sys::NGX_HTTP_SUBREQUEST_IN_MEMORY as ngx_uint_t;
+        self
+    }
```
bavshin-f5, line 139: Seems like a good place to start using `bitflags`:
```rust
bitflags::bitflags! {
    #[derive(Debug)]
    struct SubRequestFlags: ngx_uint_t {
        /// The subrequest output is stored in memory instead of being written to the client
        /// connection.
        const IN_MEMORY = nginx_sys::NGX_HTTP_SUBREQUEST_IN_MEMORY as ngx_uint_t;
        /// The subrequest's done flag is set even if the subrequest is not active when it is
        /// finalized.
        const WAITED = nginx_sys::NGX_HTTP_SUBREQUEST_WAITED as ngx_uint_t;
        // The subrequest is created as a clone of its parent. It is started at the same location
        // and proceeds from the same phase as the parent request.
        const CLONE = nginx_sys::NGX_HTTP_SUBREQUEST_CLONE as ngx_uint_t;
        // The subrequest does not block any other subrequests or the main request.
        const BACKGROUND = nginx_sys::NGX_HTTP_SUBREQUEST_BACKGROUND as ngx_uint_t;
    }
}
```
And then replace all the individual methods with `fn flags(mut self, value: SubRequestFlags) -> Self`.

src/http/mod.rs
```diff
@@ -4,6 +4,10 @@ mod request;
 mod status;
 mod upstream;
 
+/// HTTP subrequest builder and handler.
+#[cfg(feature = "alloc")]
```
bavshin-f5: The cfg is no longer needed.

src/http/subrequest.rs
```diff
@@ -0,0 +1,192 @@
...
+                self.flags as ngx_uint_t,
+            )
+        };
+        if rc != nginx_sys::NGX_OK as _ {
+            return Err(SubRequestError::Create);
+        }
+
+        let sr = unsafe { Request::from_ngx_http_request(sr_ptr) };
+        if let Some(modifier) = modifier {
+            modifier(sr).map_err(|e| SubRequestError::Modify(e))
```
bavshin-f5: Style:
```rust

if let Some(modifier) = modifier {
    modifier(sr).map_err(SubRequestError::Modify)?;
}

Ok(())
```

src/http/subrequest.rs
```diff
@@ -0,0 +1,192 @@
...
+        modifier: Option<M>,
+        handler: Option<H>,
+    ) -> Result<(), SubRequestError<E>>
+    where
+        M: FnOnce(&mut Request) -> Result<(), E>,
+        E: Display,
+        H: FnOnce(&mut Request, ngx_int_t) -> O,
+        O: IntoHandlerStatus,
+    {
+        let sr_args_ptr = self.args.as_mut().map_or(ptr::null_mut(), |args| args as *mut ngx_str_t);
```
bavshin-f5:
```rust
let sr_args_ptr = self.args.as_mut().map_or(ptr::null_mut(), ptr::from_mut);
```

src/http/subrequest.rs
ensh63: It does not work this way : (. Simple `None` is not enough; it is needed to specify a full type, like this: `None::<fn(&mut Request) -> Result<(), core::convert::Infallible>>`. Not sure if this is acceptable.

src/http/subrequest.rs
ensh63: In my opinion, individual function calls for each flag look much better than this single function with long argument.

xeioex: I prefer `flags() + bitflags` slightly more over individual methods.

src/http/subrequest.rs
ensh63: Callback definitions moved into separate methods, so no special NO_OP constants are needed now.

src/http/request.rs
```diff
@@ -104,6 +105,16 @@ where
     }
 }
 
+impl<T, E> IntoHandlerStatus for Result<T, E>
```
bavshin-f5, line 108: Please, split this change to a separate commit for visibility.

src/http/request.rs
```diff
@@ -383,62 +439,6 @@ impl Request {
...
-        // previously call of ngx_http_subrequest() would ensure that the pointer is not null
-        // anymore
-        let sr = unsafe { &mut *psr };
-
-        /*
-         * allocate fake request body to avoid attempts to read it and to make
-         * sure real body file (if already read) won't be closed by upstream
-         */
-        sr.request_body =
-            self.pool().alloc(core::mem::size_of::<ngx_http_request_body_t>()) as *mut _;
```
bavshin-f5, line 433: This might be important. Do we want to copy this part and apply it by default?

src/http/status.rs
```diff
@@ -4,23 +4,15 @@ use crate::core::Status;
...
-pub struct InvalidHTTPStatusCode {
-    _priv: (),
-}
-
-impl InvalidHTTPStatusCode {
-    fn new() -> InvalidHTTPStatusCode {
-        InvalidHTTPStatusCode { _priv: () }
-    }
-}
+pub struct InvalidHTTPStatusCode;
```
bavshin-f5: `InvalidHTTPStatusCode` constructor was private to disallow instantiating the error  outside of `ngx`. Is there a reason for changing that?

examples/subrequest.rs
```diff
@@ -0,0 +1,270 @@
...
+                    }
+                },
+            ),
+            None => {
+                if SRCtx::create(request).is_some() {
+                    let uri: &str = co.uri.to_str().unwrap_or("/proxy");
+
+                    SubRequestBuilder::new(request, uri)?
+                        .args("arg1=val1&arg2=val2")?
+                        .handler(sr_handler)
```
bavshin-f5, line 128: Would be nice to have an example for `SubRequestBuilder::init()`. Something simple, like adding a header.

examples/t/subrequest.t
```diff
@@ -0,0 +1,75 @@
...
+    }
+}
+
+EOF
+
+$t->write_file('index.html', '');
+$t->run();
+
+like(http_get('/'), 
+    qr/200 OK.*Hello from backend/s,
```
bavshin-f5: Trailing spaces.

src/http/status.rs
```diff
@@ -48,35 +40,42 @@ impl From<HTTPStatus> for ngx_uint_t {
     }
 }
 
-impl fmt::Debug for HTTPStatus {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        fmt::Debug::fmt(&self.0, f)
-    }
-}
+impl TryFrom<usize> for HTTPStatus {
```
bavshin-f5, line 51: This deserves a separate commit as well.

And maybe we should preserve `HTTPStatus::from_u16()` and implement the rest of the conversions through it.

src/http/request.rs
ensh63: There may be different situations. In some cases, it is sufficient just to discard request body. For now I suggest to allow user to make necessary adjustments in `init()` call. Another option is to add special flag to builder, but this requires additional investigation.

PR #272, opened 2026-04-03
feat: add core main-conf access

ryan-surname-p:
Add core main-conf  wrapper for accessing top-level core main
configuration in ngx-rust.

This introduces:
- CoreModuleConfExt for ngx_conf_t and ngx_cycle_t
- core_main_conf / core_main_conf_mut typed helpers
- NgxCoreModule convenience accessors for ngx_core_conf_t

The new API mirrors the existing HTTP config helper style while
centralizing the unsafe conf_ctx slot traversal in one place.

Why:

Modules that need core process configuration currently have to manually
walk cycle->conf_ctx, index with ngx_core_module.index, and cast to
ngx_core_conf_t. That is repetitive, easy to get wrong, and noisy in
reviews. 

The immediate use case is the nginx-wasm issue where module
code needs  access to core config such as the env state.

Ref: https://github.com/nginx/nginx-wasm/issues/37


ensh63:
This is a really useful addition. But I'd like to suggest to make one step further and implement configuration trait for arbitrary core module similar to `HttpModuleMainConf` with associated type for config data and functions to get it. Current approach defines interface for main core module only which may not be enough. It does not allow to define config interface for new core modules, and it does not include definitions for module-specific config data types.

ryan-surname-p:
Sure thing @ensh63 , I'll follow up next week with CoreModuleConf

ryan-surname-p:
Updated with CoreModuleMainConf, made NgxCoreModule a concrete implementation of it. 

Cleaned out the lingering public accessors from testing.

src/core/conf.rs
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    /// Get a typed mutable reference to this module's main configuration.
+    fn main_conf_mut(o: &impl CoreModuleConfExt) -> Option<&'static mut Self::MainConf> {
+        unsafe { Some(o.core_main_conf_unchecked(Self::module())?.as_mut()) }
+    }
+}
+
+/// Auxiliary structure to access `ngx_core_module` configuration.
+pub struct NgxCoreModule;
```
ensh63, line 69: I think we may expect a special trait for core module, similar to `HttpModule` trait. To make this PR minimal, I'd suggest to include only `module()` method to it for now. In future this will allow to add all other necessary components to this trait.

ryan-surname-p: Thanks for your patience and guidance on this one, I'm still wrapping my head around all this but I believe the recent change achieves what you are asking.

ensh63 (approved):

Issue #108, opened 2024-12-24
Support `stream` modules


bavshin-f5:
Currently we don't implement any wrappers for the `stream` module type. We may need that in a not-so-distant future.

PR #286, opened 2026-05-07
feat: add stream module api

plusls:
### Proposed changes
Add stream module api support

impl #108

### Checklist

### DIFF
--- a/src/stream/conf.rs
+++ b/src/stream/conf.rs
@@ -1,0 +1,267 @@
+use ::core::ptr::NonNull;
+
+use crate::{
+    ffi::{
+        ngx_conf_t, ngx_cycle_t, ngx_module_t, ngx_stream_conf_ctx_t, ngx_stream_core_srv_conf_t,
+        ngx_stream_session_t, ngx_stream_upstream_srv_conf_t,
+    },
+    stream::StreamModule,
+};
+
+/// Utility trait for types containing Stream module main configuration
+pub trait StreamModuleMainConfExt {
+    /// Get a non-null reference to the main configuration structure for Stream module
+    ///
+    /// # Safety
+    /// Caller must ensure that type `T` matches the configuration type for the specified module.
+    #[inline]
+    unsafe fn stream_main_conf_unchecked<T>(&self, _module: &ngx_module_t) -> Option<NonNull<T>> {
+        None
+    }
+}
+
+/// Utility trait for types containing Stream module server configuration
+pub trait StreamModuleServerConfExt {
+    /// Get a non-null reference to the server configuration structure for Stream module
+    ///
+    /// # Safety
+    /// Caller must ensure that type `T` matches the configuration type for the specified module.
+    #[inline]
+    unsafe fn stream_server_conf_unchecked<T>(&self, _module: &ngx_module_t) -> Option<NonNull<T>> {
+        None
+    }
+}
+
+impl StreamModuleMainConfExt for ngx_stream_conf_ctx_t {
+    #[inline]
+    unsafe fn stream_main_conf_unchecked<T>(&self, module: &ngx_module_t) -> Option<NonNull<T>> {
+        NonNull::new(unsafe { *self.main_conf.add(module.ctx_index) }.cast())
+    }
+}
+impl StreamModuleServerConfExt for ngx_stream_conf_ctx_t {
+    #[inline]
+    unsafe fn stream_server_conf_unchecked<T>(&self, module: &ngx_module_t) -> Option<NonNull<T>> {
+        NonNull::new(unsafe { *self.srv_conf.add(module.ctx_index) }.cast())
+    }
+}
+
+impl StreamModuleMainConfExt for ngx_cycle_t {
+    #[inline]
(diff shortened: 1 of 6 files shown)


src/stream/conf.rs
```diff
@@ -0,0 +1,267 @@
...
+        type MainConf = ngx_stream_core_main_conf_t;
+    }
+    unsafe impl StreamModuleServerConf for NgxStreamCoreModule {
+        type ServerConf = ngx_stream_core_srv_conf_t;
+    }
+
+    /// Stream phases in which a module can register handlers.
+    #[repr(usize)]
+    pub enum StreamPhase {
+        /// Post-accept phase
```
avahahn, line 179: these comments are not needed.

src/stream/conf.rs
```diff
@@ -0,0 +1,267 @@
...
+        if s.is_null() {
+            ngx_conf_log_error!(
+                nginx_sys::NGX_LOG_EMERG,
+                cf,
+                "failed to register {} handler",
+                S::name(),
+            );
+            return Err(AllocError);
+        }
+        // set an H::PHASE phase handler
```
avahahn, line 214: S::PHASE? also comment unneeded.

src/stream/conf.rs
```diff
@@ -0,0 +1,267 @@
...
+    }
+    unsafe impl StreamModuleMainConf for NgxStreamUpstreamModule {
+        type MainConf = ngx_stream_upstream_main_conf_t;
+    }
+    unsafe impl StreamModuleServerConf for NgxStreamUpstreamModule {
+        type ServerConf = ngx_stream_upstream_srv_conf_t;
+    }
+}
+
+pub use upstream::NgxStreamUpstreamModule;
```
avahahn, line 267: Is it possible to bundle the use statements up at the top?

src/stream/session.rs
```diff
@@ -0,0 +1,208 @@
...
+}
+
+impl Session {
+    /// Create a [`Session`] from an [`ngx_stream_session_t`].
+    ///
+    /// # Safety
+    ///
+    /// The caller has provided a valid non-null pointer to a valid `ngx_stream_session_t`
+    /// which shares the same representation as `Request`.
+    pub unsafe fn from_ngx_stream_session<'a>(r: *mut ngx_stream_session_t) -> &'a mut Session {
```
avahahn, line 118: You already added this as a From impl so I think this member function is redundant.

src/stream/session.rs
```diff
@@ -0,0 +1,208 @@
...
+        unsafe { &mut *r.cast::<Session>() }
+    }
+
+    /// Returns the result as an `Option` if it exists, otherwise `None`.
+    ///
+    /// The option wraps an ngx_stream_upstream_t instance, it will be none when the underlying NGINX
+    /// request does not have a pointer to a [`ngx_stream_upstream_t`] upstream structure.
+    pub fn upstream(&self) -> Option<*mut ngx_stream_upstream_t> {
```
avahahn, line 126: Please return a mut ref instead of a pointer here. This can be implemented as follows:
```rust
    pub fn upstream(&self) -> Option<&mut ngx_stream_upstream_t> {
        self.0.upstream.as_mut()
    }
```

This will allow rust callers to use the ref without having to play with as much unsafe code.

src/stream/session.rs
```diff
@@ -0,0 +1,208 @@
...
+        if self.0.upstream.is_null() {
+            return None;
+        }
+        Some(self.0.upstream)
+    }
+
+    /// Pointer to a [`ngx_connection_t`] client connection object.
+    ///
+    /// [`ngx_connection_t`]: https://nginx.org/en/docs/dev/development_guide.html#connection
+    pub fn connection(&self) -> *mut ngx_connection_t {
```
avahahn, line 136: Again, I would prefer this interface return a mut ref instead. Feel free to `.expect(...)` it so that you dont need to return an option either.

src/stream/session.rs
```diff
@@ -0,0 +1,208 @@
...
+        self.0.connection
+    }
+
+    /// Pointer to a [`ngx_log_t`].
+    ///
+    /// [`ngx_log_t`]: https://nginx.org/en/docs/dev/development_guide.html#logging
+    pub fn log(&self) -> *mut ngx_log_t {
```
avahahn, line 143: Same as previous comments about returning a ref.

avahahn (changes requested):
This looks good to me mostly, just a few requests.

plusls:
These methods follow the http module's `Request`, and I kept the API deliberately consistent with it. The corresponding http methods also return raw pointers, and the http module likewise keeps **both** the `from_ngx_http_request` method **and** the `From` impl:

```rust
// src/http/request.rs
pub unsafe fn from_ngx_http_request<'a>(r: *mut ngx_http_request_t) -> &'a mut Request
pub fn upstream(&self) -> Option<*mut ngx_http_upstream_t>
pub fn connection(&self) -> *mut ngx_connection_t
pub fn log(&self) -> *mut ngx_log_t
```

Returning `&mut` references would be more idiomatic Rust, but it would diverge from the existing http module style. I'd like to defer to the maintainer here: if we decide to switch to references, I'd suggest doing the same in the http module for consistency. I just followed the original style.

@avahahn

PR #291, opened 2026-05-25
feat(http): expose upstream_states and request timing/byte accessors

u5surf:
### Proposed changes

Add safe accessors for request struct fields that previously required raw-pointer deref:

  - `Request::upstream_states() -> &[UpstreamState]` walks each upstream attempt (one per peer for `proxy_next_upstream` retries) via a `#[repr(transparent)]` wrapper exposing `peer`, `status`, `response_time`, `connect_time`, `header_time`, `queue_time`, `bytes_sent`, `bytes_received`, `response_length`.
  - `Request::start_sec()`, `start_msec()`, `request_length()`, `bytes_sent()` for the request-level timing/byte fields.

`Request::upstream()` (the raw-pointer accessor) is left untouched to preserve compatibility.

Unit tests follow the `MaybeUninit::zeroed` pattern from #272.

### Checklist

### DIFF
--- a/src/http/request.rs
+++ b/src/http/request.rs
@@ -9,6 +9,7 @@ use crate::core::*;
 use crate::ffi::*;
 use crate::http::HttpPhase;
 use crate::http::status::*;
+use crate::http::upstream::UpstreamState;
 
 /// Define a static request handler.
 ///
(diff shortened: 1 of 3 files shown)


src/http/request.rs
```diff
@@ -221,6 +222,65 @@ impl Request {
...
+    /// the request line and headers.  Useful as the source label
+    /// for `ingress` byte counters.
+    pub fn request_length(&self) -> off_t {
+        self.0.request_length
+    }
+
+    /// Bytes sent to the client on this request's connection.
+    /// Reads `r->connection->sent`, the byte total nginx itself uses
+    /// to render `$bytes_sent` in `log_format`.
+    pub fn bytes_sent(&self) -> off_t {
```
avahahn, line 277: We may prefer in the future to have a separate type for surfacing `ngx_connection_t` details to the caller so It may be best to not litter request with details of it for now, but I will defer to @bavshin-f5 on that.

src/http/upstream.rs
```diff
@@ -22,3 +110,86 @@ macro_rules! http_upstream_init_peer_pt {
...
+
+    #[test]
+    fn peer_some_when_populated() {
+        let bytes = b"10.0.0.1:8080";
+        let mut peer = ngx_str_t { len: bytes.len(), data: bytes.as_ptr().cast_mut() };
+        let mut state = zeroed_state();
+        state.0.peer = &raw mut peer;
+
+        let got = state.peer().expect("peer should resolve");
+        assert_eq!(got.as_bytes(), bytes);
```
avahahn, line 156: this final assert statement is simply testing `ngx_str_t`.

src/http/request.rs
```diff
@@ -221,6 +222,65 @@ impl Request {
...
+    /// the same `ngx_http_upstream_state_t` that `u->state` points
+    /// at.
+    pub fn upstream_states(&self) -> &[UpstreamState] {
+        let arr = self.0.upstream_states;
+        if arr.is_null() {
+            return &[];
+        }
+        // SAFETY: `upstream_states`, when non-null, points to an
+        // `ngx_array_t` allocated in the request pool, valid for the
+        // lifetime of the request.
```
avahahn, line 243: I wonder if we can use a lifetime on the returned slice to inform rust that the lifetime is equivalent to the lifetime of the self reference.

src/http/upstream.rs
```diff
@@ -1,3 +1,91 @@
...
+
+impl UpstreamState {
+    /// Peer address contacted for this attempt (typically `host:port`
+    /// or `unix:/path`), or `None` when no peer was ever selected.
+    ///
+    /// `None` is the cache-HIT path where `r->upstream` exists (nginx
+    /// uses the upstream framework to consult the cache) but no
+    /// backend was contacted, plus init-time slots before peer
+    /// selection in failure cases.
+    pub fn peer(&self) -> Option<&NgxStr> {
```
avahahn, line 27: again here: I wonder if we can set the lifetime of the NgxStr to an identical lifetime attached to self.

avahahn (commented):
Just a few minor nits. Otherwise this MR does what it says it does in a relatively safe way.

u5surf:
Thanks for the review, and sorry for the delay. Rebased onto main and addressed the nits.

**`upstream.rs` test assert** — dropped; the test now just checks that `peer()` resolves a populated `ngx_str_t`, without re-asserting what `NgxStr` already covers.

**Lifetimes on the returned slice and `NgxStr`** — these are already tied to `&self` by elision. `fn upstream_states(&self) -> &[UpstreamState]` desugars to `fn upstream_states<'a>(&'a self) -> &'a [UpstreamState]`, and the same holds for `UpstreamState::peer`. Writing them out explicitly makes clippy ask for them back:

```
warning: the following explicit lifetimes could be elided: 'a
   --> src/http/request.rs:236:28
    = note: `#[warn(clippy::needless_lifetimes)]` on by default
```

which would fail CI under `RUSTFLAGS: --deny warnings`. So I have left them elided. Happy to add a doc note spelling the borrow out instead, if that would read better.

**`ngx_connection_t` details on `Request`** — leaving as is pending @bavshin-f5, since you deferred that one. Easy to move behind a separate type later if that is the direction.

`cargo fmt`, `clippy --all-targets` and the test suite are clean on nginx 1.30.4.

PR #303, opened 2026-06-24
fix: invalid context for unconfigured resolver in ngx::async_::Resolver

xeioex:
ngx_resolve_start() returns the NGX_NO_RESOLVER sentinel ((void *) -1) when the resolver has no nameservers configured. This pointer is non-null, so NonNull::new() accepted it as a valid context and it was later dereferenced and dropped via ngx_resolve_name_done(), causing invalid memory access.

### DIFF
--- a/nginx-sys/src/lib.rs
+++ b/nginx-sys/src/lib.rs
@@ -45,6 +45,13 @@ pub use stream::*;
 /// Default alignment for pool allocations.
 pub const NGX_ALIGNMENT: usize = NGX_RS_ALIGNMENT;
 
+/// Sentinel returned by `ngx_resolve_start()` when no resolver is configured.
+///
+/// nginx's `NGX_NO_RESOLVER` macro expands to `(void *) -1`, which bindgen does
+/// not emit. It is an address-only sentinel that must never be dereferenced;
+/// callers compare the `ngx_resolve_start()` return value against it.
+pub const NGX_NO_RESOLVER: *mut ngx_resolver_ctx_t = ptr::without_provenance_mut(usize::MAX);
+
 // Check if the allocations made with ngx_palloc are properly aligned.
 // If the check fails, objects allocated from `ngx_pool` can violate Rust pointer alignment
 // requirements.
--- a/src/async_/resolver.rs
+++ b/src/async_/resolver.rs
@@ -16,8 +16,8 @@ use core::ptr::NonNull;
 use core::task::{Context, Poll, Waker};
 
 use nginx_sys::{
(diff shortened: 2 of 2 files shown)


nginx-sys/src/lib.rs
```diff
@@ -45,6 +45,13 @@ pub use stream::*;
 /// Default alignment for pool allocations.
 pub const NGX_ALIGNMENT: usize = NGX_RS_ALIGNMENT;
 
+/// Sentinel returned by `ngx_resolve_start()` when no resolver is configured.
+///
+/// nginx's `NGX_NO_RESOLVER` macro expands to `(void *) -1`, which bindgen does
+/// not emit. It is an address-only sentinel that must never be dereferenced;
+/// callers compare the `ngx_resolve_start()` return value against it.
+pub const NGX_NO_RESOLVER: *mut ngx_resolver_ctx_t = ptr::without_provenance_mut(usize::MAX);
```
bavshin-f5, line 53: We have `ptr::null_mut::<T>().wrapping_offset(-1);` elsewhere, which is an equivalent way to construct `(T *) -1` without provenance. IMO, it is closer to the original definition for such constants.

bavshin-f5 (approved):
Thanks, looks correct.

I wonder if this deserves another patch release. Pretty sure it's only possible to see `NGX_NO_RESOLVER` when using or copying `clcf->resolver`, and the only available example in `ngx_http_upstream_round_robin.c` also checks `connections.nelts` at configuration time.

We might also want to check `connections.nelts` early in `Resolver::from_resolver()`. But that changes the API and can be addressed separately.

nginx-sys/src/lib.rs
xeioex: yes, I considered it

it looks a bit ugly, as we need to ensure that we have exactly `-1`.

```rust
pub const NGX_NO_RESOLVER: *mut ngx_resolver_ctx_t =
    ptr::null_mut::<core::ffi::c_char>().wrapping_offset(-1) as *mut ngx_resolver_ctx_t;
```

as `ptr::null_mut::<ngx_resolver_ctx_t>().wrapping_offset(-1)` gives `0xffffffffffffff20`

Up to you. Let me know the preferred variant.

bavshin-f5: Right, this would need `wrapping_byte_offset(-1)`.

I'm mostly worried that the conversion from `-1` to `usize::MAX` is not obvious enough, but highlighting it only makes the code uglier. Let's keep your current approach.

PR #318, opened 2026-08-28
test: enlarge the shared_dict zone for larger memory pages

u5surf:
### Proposed changes

`examples/t/shared_dict.t` declares its zone as `64k`. That is 16 slab pages where a page is 4k, but only 4 pages where it is 16k. nginx dedicates a whole page to each slot size, so storing a key and a value that fall into different size classes consumes two of those four pages, and the third `shared_dict` write in the test runs out of zone:

```
*16 http script set var handler
*16 http script var: "fst"
slab alloc: 3 slot: 0
slab alloc: 9 slot: 1
[crit] ngx_slab_alloc() failed: no memory
slab alloc: 0000000000000000
```

`ngx_http_shared_dict_set_variable` drops the write silently when the allocation fails, so the request still returns `200` and three subtests fail later on a value that was never stored:

```
Failed tests:  6-7, 10
#   Failed test 'get entries - deleted'
#          '1; fst = hello; '
#     doesn't match '(?^ms:^1; fst = new_value; $)'
```

This reproduces on Apple Silicon macOS (16k pages) and applies equally to Linux arm64 builds configured with 16k or 64k pages. Only the 4k page systems CI currently runs on have enough room, which is why it has gone unnoticed.

`512k` keeps the zone at eight pages or more on 4k, 16k and 64k page systems alike, matching the minimum that nginx's own zone-backed modules enforce (`8 * ngx_pagesize`).

### Testing

`examples/t/shared_dict.t` against nginx 1.30.4 built from source on Apple Silicon (16k pages):

| zone | result |
| --- | --- |
| `64k` (current) | `Failed 3/14 subtests` |
| `256k` | `All tests successful` — but only 4 pages on a 64k page system |
| `512k` (this PR) | `All tests successful` |

I do not have a 4k page machine to hand, so I have not run it there; the change only grows the zone, and CI covers that configuration.

### Checklist

### DIFF
--- a/examples/t/shared_dict.t
+++ b/examples/t/shared_dict.t
@@ -36,7 +36,7 @@ events {
 http {
     %%TEST_GLOBALS_HTTP%%
 
-    shared_dict_zone z 64k;
+    shared_dict_zone z 512k;
     shared_dict $arg_key $foo;
 
     server {


bavshin-f5 (approved):
Looks good, thanks!

8 pages might be insufficient in general, as the input is user-controlled and can be placed into different slots of the slab allocator. However, it's good enough for this test, as verified in my Linux environment with 64k pages.

> One caveat from those runs: t/shared_dict.t subtest 3 ("check value 2") failed intermittently, roughly one run in four, independently of the zone size. It looks like a worker_processes 2 timing issue in the test itself rather than anything in the module, but it is worth knowing before wiring the suite into another job.

Unfortunately, I can't reproduce the problem. However, it might be caused by uneven request distribution between workers.
Is it limited to a specific environment? Can you share logs of a failed run with `env TEST_NGINX_CATLOG=1 TEST_NGINX_VERBOSE=1 prove -v ...`?

u5surf:
Logs of failing runs, before and after this PR:
https://gist.github.com/u5surf/544c3d9bd1746f1ee48ba2a5fcc30a0c

macOS 14.6, Apple Silicon, 16k pages, 8 cores, nginx 1.30.4.

### Why you could not reproduce it

The flags make it go away. Same revision, same machine, 15 runs each:

| | failures |
| --- | --- |
| `prove shared_dict.t` | 14 / 15 |
| `env TEST_NGINX_CATLOG=1 TEST_NGINX_VERBOSE=1 prove -v shared_dict.t` | 1 / 15 |

Printing every request and response puts enough of a gap between requests to
change which worker accepts the next one. So the attached logs are two lucky
captures under `-v`; without it, it fails almost every run.

### What the two logs show

| | failing subtests | `ngx_slab_alloc() failed` |
| --- | --- | --- |
| before, 64k zone | 3, **6-7, 10** | 1 |
| after, 512k zone | 3-4 | 0 |

Subtests 7 and 10 read `$shared_dict_entries`, so they allocate a string sized
for every entry at once. Those are the ones this PR fixed, and they are clean
after it.

What is left is 3, 4 and 6, the three `check()` calls. They fail before this PR
as well, so they are a separate and pre-existing problem, and they are what I
flagged. In the second log, 67 of the 70 responses come from one worker, so
`check()` exhausts its 25 retries without ever seeing a second pid.

### Is it environment specific

I think so, and checking needs neither the module nor the test suite:

```nginx
worker_processes 2;
events { }
http {
    server {
        listen 127.0.0.1:18099;
        location / { return 200 "$pid\n"; }
    }
}
```

```console
$ for i in $(seq 1 40); do curl -sS http://127.0.0.1:18099/; done | sort | uniq -c
  2 52118
 38 52120
```

If that comes out near 20/20 on your machine, that is the whole difference
between our environments.

### A possible fix

Pausing between retries, so the worker that just answered goes back to sleep,
does deliberately what `-v` was doing by accident:

| `check()` | failures |
| --- | --- |
| as merged | 20 / 20 |
| 0.2s between retries | 1 / 50 |
| 0.2s, retry ceiling 25 → 60 | 0 / 30 |

Costs about 0.6s where the first retry already lands on the other worker, since
the loop returns as soon as it sees a second pid. Happy to open a PR if that
looks like the right shape.

PR #319, opened 2026-08-29
fix(examples): skip the awssig example on Darwin

u5surf:
### Proposed changes

Building the examples as nginx modules fails to link on macOS. The `awssig`
example depends on `chrono`, which reads the local timezone through
`iana-time-zone`, which resolves it through CoreFoundation on Apple platforms:

```
"_CFStringGetCStringPtr", referenced from:
    iana_time_zone::platform::string_ref::StringRef$LT$T$GT$::as_utf8
    in libawssig.a[223](iana_time_zone-....o)
"_CFTimeZoneCopySystem", referenced from:
    iana_time_zone::platform::system_time_zone::SystemTimeZone::new
    in libawssig.a[223](iana_time_zone-....o)
ld: symbol(s) not found for architecture arm64
```

Putting CoreFoundation on the link line is not an option: it is not fork-safe,
and nginx forks its workers from the master process. So this skips the module
on Darwin instead, the way `orig_dst` is already limited to Linux.

CI does not currently notice any of this, because the macOS job in `ci.yaml`
runs `cargo build`/`cargo test` and never configures nginx with `--add-module`.
Anyone following the examples' README on a Mac hits it immediately.

### Testing

`auto/configure --add-dynamic-module=.../examples && make` against nginx 1.30.4
on Apple Silicon:

| | before | after |
| --- | --- | --- |
| configure | five example modules | four, without awssig |
| link | `ld: symbol(s) not found for architecture arm64` | succeeds |

The four remaining examples build and link unchanged. No other platform is
affected.

### Checklist


bavshin-f5:
CoreFoundation is not fork-safe and thus can not be linked to nginx. You need to skip this module on macOS instead.

u5surf:
Thanks, that makes sense. Reworked to skip the module on Darwin instead of linking CoreFoundation, following the `orig_dst` precedent, and rebased onto main.

Verified on macOS: `configure` now emits four example modules without awssig, and `make` links them cleanly.

bavshin-f5 (approved):
Thanks!

PR #328, opened 2026-09-14
Update Bindgen

avahahn:
This commit updates bindgen to latest tagged version, thus avoiding the [incorrect bitfield layout issue](https://github.com/rust-lang/rust-bindgen/issues/3384) that was recently fixed.


bavshin-f5 (approved):
Thanks!

PR #332, opened 2026-09-17
test: make the example tests pass on macOS

u5surf:
### Proposed changes

Two unrelated failures in `examples/t`, both only visible once the functional
suite runs on a Mac, which no CI job does yet. #320 adds such a job and both
turned up there.

**`awssig.t`** writes a config using a module that is no longer built on
Darwin after #319, so nginx refuses to start. Skip the file there, matching the
configure-time condition.

**`shared_dict.t`** fails on its `check()` calls, subtests 3, 4 and 6. `check()`
records the worker pid that answered and retries back to back until a different
one answers, which is what proves the value came from shared memory rather than
worker-local state. Consecutive connections keep landing on the worker that is
already awake, so the loop exhausts without ever seeing a second pid. With
plain nginx and two workers:

```nginx
worker_processes 2;
events { }
http {
    server {
        listen 127.0.0.1:18099;
        location / { return 200 "$pid\n"; }
    }
}
```

```console
$ for i in $(seq 1 40); do curl -sS http://127.0.0.1:18099/; done | sort | uniq -c
  2 52118
 38 52120
```

Pausing between retries lets the worker that just answered go back to sleep, so
the kernel has a reason to wake the other one.

### Testing

macOS 14.6, Apple Silicon, nginx 1.30.4, running `shared_dict.t` in a loop:

| | failures |
| --- | --- |
| before | 50 / 50 |
| after | 0 / 50 |

The loop returns as soon as it sees a second pid, so on a platform where the
first retry already lands on the other worker the cost is one pause per
`check()`. The whole of `examples/t` takes about 2s here either way.

`accept_mutex on` and `listen ... reuseport` were both tried first; on this
platform each made the split worse, 40 / 0.

### Checklist


u5surf:
Folded into #320, which is where both failures surfaced.

